PR #103, opened 2015-01-12
Fix #102 - Add case sensitivity.

### LINKED ISSUE #102: Add case sensitivity.
On my i3 configuration I tend to symlink common programs to Title-cased names in my personal bin folder. For example,

    Terminal  ->  gnome-terminal -e screen
    Chrome    ->  google-chrome

This way, given that with dmenu/dmenu_run the search is case sensitive, it's quite easy to filter out rarely used programs (since most have lower-case names).

It would be nice to have case sensitivity support in rofi.

EdwinPujols:
This adds case sensitivity support.

### DIFF
--- a/config/config.def.c
+++ b/config/config.def.c
@@ -99,6 +99,8 @@ Settings config = {
     .disable_history   = FALSE,
     /** Use levenshtein sorting when matching */
     .levenshtein_sort  = FALSE,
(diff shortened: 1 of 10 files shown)


EdwinPujols:
Hi @DaveDavenport. The last commit contains the Xresources option and the visual indicator (a very simple visual indicator).

PR #105, opened 2015-01-14
SSH config file parser: Recognize multiple host names in a "Host" line

Tblue:
Currently, rofi's SSH config file parser only recognizes a *single* host name in a `Host` line when parsing the user's SSH config file. Consequently, the following block yields a single host name, "milliways":

```
Host milliways slartibartfast slarti
    Port 1234
```

However, rofi should present all three host names to the user, i. e. "milliways", "slartibartfast" and "slarti".

This patch modifies the config file parser so that it correctly recognizes multiple host names in a `Host` line. Additionally, it also makes the parser ignore comment lines and negated patterns.

### DIFF
--- a/source/ssh-dialog.c
+++ b/source/ssh-dialog.c
@@ -49,6 +49,10 @@
 
 #define SSH_CACHE_FILE    "rofi-2.sshcache"
 
+// Used in get_ssh() when splitting lines from the user's
+// SSH config file into tokens.
+#define SSH_TOKEN_DELIM   "= \t\r\n"
+
 static inline int execshssh ( const char *host )
 {
     /**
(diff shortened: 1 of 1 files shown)


DaveDavenport:
You are right.  Patch looks good on first glance, one minor suggestion: Can you change str(n)dup to g_str(n)dup.  The non-glib strndup is not always available.

Tblue:
Done. I also changed `strcasecmp()` to `g_ascii_strcasecmp()` since if I remember correctly, the former is not always available either. Using glib's ASCII version is probably okay, since host names usually are plain ASCII.

DaveDavenport:
If people complain we can always switch to g_utf8_collate.

I will merge the patch this evening.

Tblue:
All right. Thanks!

PR #109, opened 2015-01-14
Mouse navigation: Avoid unnecessary looping.

EdwinPujols:
Since the `i` counter is increasing, the if-block should just break from the loop right there.


DaveDavenport:
thx.

PR #117, opened 2015-01-28
typo in install docs

geier:


DaveDavenport:
Read the manual carefully, the `../configure` is correct for the given instructions.
The way explained in the install instructions builds the source in a separate directory (named build) and therefor does not pollute the source directory.  The actual configure script is in the parent directory.

DaveDavenport:
owh there is an error..  I forgot the cd build..  my bad :(
Will fix that.

geier:
Oh, looks like I didn't think about *why* the build directory is created (and didn't notice that it stayed empty)...

DaveDavenport:
Thanks for reporting, it is good to have these things squashed. I am pretty bad at this. Added the `cd build` to the install guide.

PR #127, opened 2015-02-10
Modified the event delay to make rofi more usable

lmbrt:
I use rofi as an application launcher on a daily basis.  The 200ms delay that was introduced has severely hindered the usefulness of the tool because it launches the incorrect command unless I wait for the screen to update.

Basic workflow:

* Launch rofi
* Enter command, i.e. firefox
* Hit enter

At 200ms, rofi will launch a historical command (i.e. pidgin) unless I wait for it to update and display firefox.  I've tried training myself for several days to wait, but it's consistently happening.

Changing the 200ms to 50ms seems to alleviate the issue for me.  I tried 100ms as well, but preferred 50ms at the end of the day.  Now the screen updates as fast as I can type.

### DIFF
--- a/source/rofi.c
+++ b/source/rofi.c
@@ -190,8 +190,8 @@ static inline MainLoopEvent wait_for_xevent_or_timeout ( Display *display, int x
     FD_ZERO ( &in_fds );
     FD_SET ( x11_fd, &in_fds );
 
-    // Set our timer.  200ms is a decent delay
-    tv.tv_usec = 200000;
+    // Set our timer.  50ms is a decent delay
+    tv.tv_usec = 50000;
     tv.tv_sec  = 0;
     // Wait for X Event or a Timer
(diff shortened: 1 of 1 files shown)


DaveDavenport:
The lazy loading should only be enabled if you have more then 5000 entries.
You can also trying to raise this limit.

lmbrt:
Increasing the lazy-filter-limit does seem to accomplish the same goal.  I missed this option / didn't realize this was causing the delay.

Could consider disabling lazy-filter-limit if it's set to 0

DaveDavenport:
Set it to -1 will do that (up to 4294967295 items).

The lazy delay was added because on large lists and slow machine, constantly filtering caused a big 'jitter'. 

I did however never expected somebody to have 5000+ executables in its path :D

lmbrt:
Thank you for the -1 value.  I've updated my environment and am much happier now.

No more "waiting" to launch my command.

DaveDavenport:
Good.

PR #129, opened 2015-02-16
Fix bugs relating to hotkeys with modifiers

gbitzes:
This commit makes the following changes:
1) Fix a bad check introduced in dee0bfb, which checks the modstate of the
event with the keysym of the switcher, not its modmask:

        modstate & ( switchers[i].keysym )

I think it was broken before as well, a binary AND between the two masks
will not tell much about whether they are equal. Instead, the masks are
now checked for equality after removing LockMask and NumlockMask.

2) Stop binding to AnyModifier when none is used. This conflicts with the case
in which we want to bind different modes to the same key, but different
modifiers.

for example, rofi -key-run F2 -key-ssh alt+F2 -key-window shift+F2

Binding F2 with AnyModifier would prevent alt+F2 and shift+F2 from being used.

3) Remove UngrabKey from x11_grab_key to make the above example possible.
4) Refactor search for any of the global keybindings into function locate_switcher.

### DIFF
--- a/source/rofi.c
+++ b/source/rofi.c
@@ -544,6 +544,27 @@ inline static void menu_nav_down ( MenuState *state )
     state->update = TRUE;
 }
 /**
+ * @param key the Key to match
+ * @param modstate the modifier state to match
+ *
+ * Return the index of the switcher that matches the key combination
+ * specified by key and modstate. Returns -1 if none was found
+ */
+extern unsigned int NumlockMask;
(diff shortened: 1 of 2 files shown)


DaveDavenport:
he he , the keysym with modstate is a silly one.
The AnyModifier is a left over from simpleswitcher, but it seems sane to change.  (I never use Rofi in daemon mode, so haven't realy looked at this code.

Thanks for the fixes.

PR #132, opened 2015-02-27
add missing automake macro

mbakke:
Rofi 0.15.2 fails to compile on my system unless this macro is included. Didn't check where it came from.


DaveDavenport:
What version of automake are you on?
manual sais:
```
AM_PROG_CC_C_O

    This is an obsolescent macro that checks that the C compiler supports the -c and -o options together. Note that, since Automake 1.14, the AC_PROG_CC is rewritten to implement such checks itself, and thus the explicit use of AM_PROG_CC_C_O should no longer be required.
```

mbakke:
That was fast. It seems my distro uses 1.12, will try with 1.14 and update accordingly. Thanks!

DaveDavenport:
maybe I need to apply this one anyway. I guess more people will hit it.

mbakke:
Can confirm 1.14 works fine. Probably sufficient to list >=automake-1.14 as a dependency, I suppose most distros carry multiple autotools versions anyway (1.14 is almost two years by now).

DaveDavenport:
I have fixed this in git..  Updated INSTALL.md and set it in the Makefile.am

roomcays:
Just to note, that current Gentoo does not provide automake-1.14, most recent stable is 1.13.4. This leads to an error coming from `autoreconf --install` and newest rofi being unable to compile of course. 

Following patch makes it compile again:
```diff
diff --git Makefile.am Makefile.am
index 33b9582..8e22d90 100644
--- Makefile.am
+++ Makefile.am
@@ -1,5 +1,5 @@
 # Specify automake version.
-AUTOMAKE_OPTIONS  = 1.14.0
+AUTOMAKE_OPTIONS  = 1.13.4
 
 ##
 # Rofi the program
diff --git configure.ac configure.ac
index 3c6445d..834ebe6 100644
--- configure.ac
+++ configure.ac
@@ -14,6 +14,7 @@ AM_SILENT_RULES([yes])
 ##
 AC_PROG_CC([clang gcc cc])
 AC_PROG_CC_C99
+AM_PROG_CC_C_O
 
 AC_USE_SYSTEM_EXTENSIONS
 
```

Hopefully 1.14 will be unmasked soon...

PR #136, opened 2015-03-03
xrmoptions: give a name to the union in XrmOption struct

jpleau:
fixes a compilation warning when building with -Wpedantic


DaveDavenport:
Thanks

PR #139, opened 2015-03-30
update documentation to new syntax

simonsmiley:
"-rnow" is now deprecated. "-show run" is the new way to go.


DaveDavenport:
thanks.

PR #145, opened 2015-04-08
manpage: add examples for dmenu

jpleau:


bcremer:
:+1: 

Never really understood this functionality before.

DaveDavenport:
Only this:
```
echo "Option #1\nOption #2\nOption #3" | rofi -dmenu
```
Under bash  does not work. This does:
```
echo -e "Option #1\nOption #2\nOption #3" | rofi -dmenu
```

jpleau:
Oops. Fixed.

Worked in zsh, I always forget about -e

PR #172, opened 2015-05-26
Fix font bug, use pango_layout_get_size to determine height

gbitzes:
This commit fixes a bug introduced in 9530f195d1055da8dffbd40d6f3c2f7c73749598, it appears that the information provided by `textbox_get_estimated_char_height` is not correct. Using `Source Code Pro` for example, no matter how small or big I set the font size to be, it always returns the same value: `19456`

The end result is that a large font size produces text that is vertically chopped off, try using ``Mono 20`` to see an example.

I am not a pango expert, but it looks like ``pango_layout_get_size`` is what we want here to get the font height, which is what this commit changes.

Cheers

### DIFF
--- a/source/textbox.c
+++ b/source/textbox.c
@@ -816,11 +816,10 @@ int textbox_get_estimated_char_height ( void )
     pango_layout_set_font_description ( layout, pfd );
     pango_font_description_free ( pfd );
 
-    // Get width
+    // Get height
     PangoContext     *context = pango_layout_get_context ( layout );
-    PangoFontMetrics *metric  = pango_context_get_metrics ( context, NULL, NULL );
-    int              height   = pango_font_metrics_get_ascent ( metric ) + pango_font_metrics_get_descent ( metric );
-    pango_font_metrics_unref ( metric );
+    int              height;
+    pango_layout_get_size( layout, NULL, &height );
 
     g_object_unref ( layout );
     return ( height ) / PANGO_SCALE + 2 * SIDE_MARGIN;


DaveDavenport:
```c
pango_layout_get_size()
```
Is not correct, it gets the size of the current string, the actual possible height is therefor not taken into account.    (e.g. string `aaa`  will be less high then `aap` and that will be less heigh then `laap`. ) You need to take the propper ascent and descent into account.
This picture explains  it to me :smile: :


I do see that something broke though, and it is currently returning me always 16+4. Not sure when/why this broke. I know that at the time of making this change I tested it and it was correct...

DaveDavenport:
```patch
diff --git a/source/textbox.c b/source/textbox.c
index 81a9988..9bc37e2 100644
--- a/source/textbox.c
+++ b/source/textbox.c
@@ -809,19 +809,14 @@ double textbox_get_estimated_char_width ( void )
 
 int textbox_get_estimated_char_height ( void )
 {
-    // Create a temp layout with right font.
-    PangoLayout          *layout = pango_layout_new ( p_context );
     // Set font.
     PangoFontDescription *pfd = pango_font_description_from_string ( config.menu_font );
-    pango_layout_set_font_description ( layout, pfd );
-    pango_font_description_free ( pfd );
 
     // Get width
-    PangoContext     *context = pango_layout_get_context ( layout );
-    PangoFontMetrics *metric  = pango_context_get_metrics ( context, NULL, NULL );
-    int              height   = pango_font_metrics_get_ascent ( metric ) + pango_font_metrics_get_descent ( metric );
+    PangoFontMetrics *metric = pango_context_get_metrics ( p_context, pfd, NULL );
+    int              height  = pango_font_metrics_get_ascent ( metric ) + pango_font_metrics_get_descent ( metric );
     pango_font_metrics_unref ( metric );
 
-    g_object_unref ( layout );
+    pango_font_description_free ( pfd );
     return ( height ) / PANGO_SCALE + 2 * SIDE_MARGIN;
 }
```

Can you try if the above patch fixes it?

DaveDavenport:
I have committed the patch, as it seem to be cleaner in any case.

gbitzes:
Yes, it fixes the issue. Thanks!

PR #195, opened 2015-07-28
Remove trailing whitespaces

jubalh:


DaveDavenport:
Aah it seems make indent does not take these into account..  that needs to be fixed!

PR #201, opened 2015-08-03
Replace x-terminal-emulator with xdg-terminal

jubalh:
x-terminal-emulator is working only on debian and derivates AFAIK. I
think xdg-open is the correct thing to use here since it behaves the
same across all distros.

However I am not 100% sure if my change is correct. Please test it too.

### DIFF
--- a/config/config.def.c
+++ b/config/config.def.c
@@ -78,16 +78,16 @@ Settings config = {
     .menu_hlbg_urgent = "#DC322F",
     .menu_hlbg_active = "#268BD2",
     /** Terminal to use. (for ssh and open in terminal) */
-    .terminal_emulator = "x-terminal-emulator",
+    .terminal_emulator = "xdg-terminal",
     .ssh_client        = "ssh",
     /** Command when executing ssh. */
-    .ssh_command       = "{terminal} -e {ssh-client} {host}",
+    .ssh_command       = "{terminal} '{ssh-client} {host}'",
     /** Command when running */
     .run_command       = "{cmd}",
     /** Command used to list executable commands. empty -> internal */
     .run_list_command  = "",
     /** Command executed when running application in terminal */
-    .run_shell_command = "{terminal} -e {cmd}",
+    .run_shell_command = "{terminal} '{cmd}'",
     /**
      * Location of the window.
      * Enumeration indicating location or gravity of window.
--- a/doc/rofi-manpage.markdown
+++ b/doc/rofi-manpage.markdown
@@ -409,7 +409,7 @@ The following options are further explained in the theming section:
       rofi -terminal xterm
 
   Pattern: *{terminal}*
-  Default: *x-terminal-emulator*
+  Default: *xdg-terminal*
 
 `-ssh-client` *client*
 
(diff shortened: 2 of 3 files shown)


DaveDavenport:
```xdg-terminal``` does not exist on my system, or is available in a package.

jubalh:
And what system is that @DaveDavenport ?

jubalh:
Also this seems to indicate to try to xdg-termianl first:
http://unix.stackexchange.com/questions/137782/launching-a-terminal-emulator-without-knowing-which-ones-are-installed

AFAIK it should be the distribution unspecific way of launching the default terminal.
x-terminal-emulator will only work on debian based ones AFAIK.

carnager:
@jubalh there is no distribution agnostic way to call a terminal. x-terminal-emulator only works on debian. Maybe on fedora too, I don't know. And as for xdg-terminal... I haven't found any major distribution that ships with it. Fedora doesn't. Ubuntu/Debian don't. Arch Linux: doesn't.

DaveDavenport:
I do not seem to have it on debian,  apt-file does not find it in any package.
According to @carnager  it also does not exists on archlinux and fedora.

maybe the distro's packages should update the default in rofi, I could make it a configure option.
I don't realy see it as rofi's task to figure this out. (we could provide a script that does the what the stackexchange page suggest?)

DaveDavenport:
I went with the i3 method, rofi now includes a rofi-sensible-terminal script that tries xdg-terminal, x-terminal-emulator and several others.

jubalh:
@DaveDavenport This is a great solution, thanks a lot!
With commits da044e0a055b0264e0d99efd9fe35004242999bc and 7e9ac526b54ce3b35ad292a4e091d25d75e47eef I can continue packaging rofi.

DaveDavenport:
New release after my holiday. So end of august.

jubalh:
Alright

jubalh:
Unfortunately there is some mistake in commit da044e0a055b0264e0d99efd9fe35004242999bc
I created a comment there. xdg-terminal does not support the -e option.

DaveDavenport:
aah F!#!@# stupid incompatibility.

DaveDavenport:
Ill remove xdg-terminal for now.. it seems not to be in http://portland.freedesktop.org/xdg-utils-1.0/  and  not in the upcoming 1.1 release.

jubalh:
I will contact the authors and freedesktop to find out more.

jubalh:
No response from them yet.
On your distribution you probably have to install xdg-utils to have xdg-open. Also I found other projects use it, for example: https://github.com/thestinger/termite/

DaveDavenport:
We have a `rofi-sensible-terminal` now.

jubalh:
Which still just opens the first terminal from a list it can find. Compared to xdg-terminal which tries to open the most suitable terminal for the environment. So when you are on XFCE xdg-terminal opens xfce4-terminal. On KDE it opens konsole.

DaveDavenport:
Yes, issue stays..  maybe make a merge request that adds xdg-terminal as first in list of rofi-sensible-terminal

DaveDavenport:
let me fix that

DaveDavenport:
done

jubalh:
Just for reference:
It was added with commit d01afc53032b52c07bdf33330f77701534305eea

PR #203, opened 2015-08-04
Call bash instead of sh for rofi-sensible-terminal

jubalh:
In the script you use 'command -v' which AFAIK is a bash built-in and
thus not available on all shells. Although /bin/sh links to /bin/bash on
many systems it does not on every system. So it is possible that it does
not work.


DaveDavenport:
oeps you are right..  copied it from  i3-sensible-terminal :smile:

PR #208, opened 2015-08-15
Fix argument quoting when executing shell commands

svenstaro:
We need this because some shells (such as termite) will not consider everything behind -e as an argument to the -e command but as their own. To  be safe, -e should be in the format: -e "everything for the command in here" as opposed to -e everything -out -here -instead because shells might try to evaluate -out and -here for their own arguments. This fix was tested with xterm and termite.

Before this fix, using termite as a terminal didn't yield expected results.

### DIFF
--- a/config/config.def.c
+++ b/config/config.def.c
@@ -81,13 +81,13 @@ Settings config = {
     .terminal_emulator = "rofi-sensible-terminal",
     .ssh_client        = "ssh",
     /** Command when executing ssh. */
-    .ssh_command       = "{terminal} -e {ssh-client} {host}",
+    .ssh_command       = "{terminal} -e '{ssh-client} {host}'",
     /** Command when running */
     .run_command       = "{cmd}",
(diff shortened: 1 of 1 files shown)


DaveDavenport:
I remember from earlier experiments with this that quoting breaks rxvt(-unicode) and and some others.

svenstaro:
Well, currently it breaks termite. :-/

DaveDavenport:
non standard argument handling sucks yes. Again this is why I made this a configurable option.

xterm does:
```
       -e program [ arguments ... ]
               This option specifies the program (and its command line arguments) to be run in the xterm window.  It also sets the window title and icon name to be the basename of the program being executed if neither -T nor  -n  are
               given on the command line.  This must be the last option on the command line.
```

On debian if a terminal does not provide this behavior they provides a wrapper. For example `gnome-terminal.wrapper`. 

I would like to have a 'unified' fix for this, but fixing one will break another. For now I will stick to the current behavior.

PR #209, opened 2015-08-17
Monitor -2 stands for the focused window

t-wissmann:
When setting rofi's monitor to -2, rofi is placed over the focused
window.

This is in particular useful for window-related scripts, e.g. actions on that menu (moving to different desktop) or menus that offer some further way of text input.


DaveDavenport:
thx

PR #210, opened 2015-08-18
Make lazy filter timeout configurable and enable continous script mode

stettberger:


DaveDavenport:
Can this one be split into 2 patches? one for the timeout and one for the continuous script mode.

Can you also explain what continuous mode does?

PR #221, opened 2015-09-04
First compute cmd_list length before passing it

t-wissmann:
In switcher_run(), first compute the cmd_list and in particular its
length before passing it to menu().

In C the evaluation order of parameters to a function call is
unspecified, see e.g. http://en.cppreference.com/w/c/language/eval_order
for some background. So it might happen (and indeed did happen on my
machine) that cmd_list_length (of value 0) is pushed to the stack before
sw->get_data() sets the variable to the actual length.


DaveDavenport:
ugh, yes thanks.. that wasn't to smart :frowning:.
I am very happy people are code reviewing rofi to catch these stupid mistakes.

PR #235, opened 2015-09-28
Store command in history only if execution is successfull

koppa:
Currently commands that cannot  be executed are stored in the history.

### DIFF
--- a/source/dialogs/run.c
+++ b/source/dialogs/run.c
@@ -37,6 +37,7 @@
 #include <strings.h>
 #include <string.h>
 #include <errno.h>
+#include <stdbool.h>
 
 #include "rofi.h"
 #include "helper.h"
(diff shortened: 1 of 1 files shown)


DaveDavenport:
Sounds like a sane thing to do.
I think we should add this to the ssh 'view' as it also uses the history?
And lets use integer for now instead of `bool`. If we switch to using `stdbool.h` (not a bad idea) we should directly do this in throughout rofi.

Thanks.

koppa:
I will fix the commit later today.

The implementation for the ssh dialog is not straight forward.
You have to check the result of the *ssh* process.
This would not be possible in the one shot mode, or needs an additional thread.

DaveDavenport:
True, ssh is a bit trickier in current form.

DaveDavenport:
fixed thanks.

PR #236, opened 2015-09-29
Only cache successful runs

daemonl:
I keep mis-typing names of commands, and they get cached which makes them show up next time.
This only caches if no error was thrown.


DaveDavenport:
fixed thanks.

PR #238, opened 2015-09-30
Filtering speed improvements

larkery:
I have been using rofi in place of dmenu, it's very nice but it's quite a lot slower for really long lists of inputs (hence the lazy filter limit I guess).

Here are two commits which make it a bit more efficient; the first one avoids refiltering the whole list when the input becomes more restrictive (pretty common as you normally add characters on the end). 

The second avoids repeatedly duplicating and case-folding the input items array every time it is refiltered. This makes some of the code slightly uglier (more state to manage), but it makes it significantly faster.

I think I have avoided introducing any bugs, but it might be worth checking that I am freeing everything I should be at the relevant places. The additional state is in `struct MenuState` in `rofi.c`, fields `last_matched_tokens` and `case_folded_lines`.

### DIFF
--- a/doc/rofi-manpage.markdown
+++ b/doc/rofi-manpage.markdown
@@ -323,7 +323,7 @@ Set the spacing between the rows.
 
 `-separator-style` *style*
 
-Set separator style. Possible options are "solid" or "dash".
+Set separator style. Possible options are "none", "solid" or "dash".
 
     Default: *dash*
 
--- a/doc/rofi.1
+++ b/doc/rofi.1
@@ -544,7 +544,7 @@ Max:     *50*
 \fB\-separator\-style\fR \fIstyle\fR
 .
 .P
-Set separator style\. Possible options are "solid" or "dash"\.
+Set separator style\. Possible options are "none", "solid" or "dash"\.
 .
 .IP "" 4
 .
@@ -967,7 +967,7 @@ Add a message line below the filter entry box\. Supports pango markup\. For more
 \fB\-normal\-window\fR
 .
 .P
-Make \fBRofi\fR reacts like a normal application window\. Useful for scripts like Clerk that are basically an application\.
+Make \fBrofi\fR reacts like a normal application window\. Useful for scripts like Clerk that are basically an application\.
 .
 .SS "Message dialog"
 \fB\-e\fR \fImessage\fR


larkery:
Not sure whether my commit has caused the CI build to fail :/

DaveDavenport:
Some of the tests are failing, I can look at this.

So what are the costs of this patch? I guess we increase in memory as we need to keep string+(casefolded and collated) copy around and the initial startup time (that is normally an unfiltered list) goes up because the casefolding has to be done.

larkery:
Memory use for a second copy of each string, which is (I think) unavoidable. 

Startup speed is affected at the moment, but taking out line 1102 of `rofi.c` would remove that; the case-folded text is regenerated in the filtering function if it doesn't already exist so it will work either way.

Having it come up quickly and then take a second to do the first filter is probably better than having it come up slowly and do the first filter more quickly, so I'll change that now.

DaveDavenport:
and what about switching between case sensitive and insensitve at runtime? (` key)

larkery:
Oh, and there is a small memory cost for the first patch, in that two sets of tokenized inputs may be around, but that is not likely to be significant.

I've done a little timing and for some of the inputs I use (lists of ~400,000 files) the case-folding takes on the order of a second to run (1-2s typically) and the remainder of the filtering around 40 microseconds, so it does make a perceptible difference in those cases.

DaveDavenport:
What is more expensive,casefolding or collating? maybe making a 'ascii' mode is interesting? we spend alot of cpu now incase we have utf8.

I am all for speeding things up, startup time and 'response'  time..  We could make the reading/filtering async, but afraid that you get annoying things that typing starts to run ahead of what is shown and therefor <type><enter> actually activates wrong result. (windows 8.1 search and atom editor quick panel have a habit of doing this, and its horrible).


(second thing? do we really want to support showing 400.000 lines? something about right tool for the right job?)

DaveDavenport:
@carnager  With Clerk do you have an opinion?

DaveDavenport:
I like the incremental filtering, I think this would be a very good addition and should be applied anyway.
I am not sure about moving work to startup time and keeping 2 copies around. (well you need three if you want to quickly switch between case sensitive and insensitive).

I **do not** and **will not** accept patches that introduce `goto`'s.

larkery:
The ` key is still supported; I tweaked the handler to recompute the pre-folded items. I agree that 400K lines is quite a lot; that is the list of non-dotfiles in my home directory. I use it as a quick way to get to files whose path I forget, and I do find it useful to be able to do that in an interactive way. It is performant in normal dmenu, which I could happily use for that purpose if you would rather rofi was not tuned for that kind of size of input.

Casefolding and collating are the same for the purposes of this patch - I should have named some of the functions better. I would expect them to have about the same runtime, as I guess the bulk of it is in doing malloc + memcpy. 

I can imagine another option involving collating the input list once at the start, and then using a unicode aware variation of strcasecmp / strcmp in the matching code, rather than downcasing the values and using strcmp. This might be neater overall?

I'll take out the `goto` - I know they are controversial (personally I think they are fine for breaking out of a specific loop, as the alternative is making a bit of state and poking at it which I find slightly harder to reason about, but it's just an opinion, man :) ) 

The code as it stands now does not increase time to get to any point (unless you end up having to swap because of the second array of things), as it does the first case fold + collate at the same point as the master branch would (on the first filter), and then simply avoids redoing it for all the subsequent filters. I expect it will be a little quicker even to the first filtering, because it is not allocating and deallocating in a loop, but that's probably marginal.

DaveDavenport:
The goto is more a personal thing.. At work I never let students/co workers write goto's because of the obvious reasons (abuse most of the case).  So I don't write goto's myself or have them in my apps. (even embedded apps.)

DaveDavenport:
ugh  I broke CI.. should be fixed now.

carnager:
i tried the fork and from a user's point of view it seems to be fine. definitely less laggy

larkery:
I'll split the two changes into two separate PRs as they are kind of orthogonal anyway.
-----------------------

PR #240, opened 2015-10-01
Improve filtering efficiency

larkery:
If the previous query is a strict superset of the current query then we
only need to do matching on the previous result set, rather than on all
input lines.

This is just the part about incremental filtering (with goto gone!)

### DIFF
--- a/source/rofi.c
+++ b/source/rofi.c
@@ -216,6 +216,7 @@ typedef struct MenuState
     int          refilter;
     int          rchanged;
     int          cur_page;
+    char         **last_matched_tokens;
 
     // Entries
     textbox      *text;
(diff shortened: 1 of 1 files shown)


DaveDavenport:
I see some todo's in the comments? what are the things we need to look at? Just fuzzy matching?

(Thanks for all your effort :smile: )

larkery:
Oh, I have turned it off for fuzzy matching because I did not have enough spare brainpower last night to determine whether the method is correct for fuzzy matching. My guess is that it does work, but didn't want to chance it. The todo is to think about whether to remove the test in the if, in which case fuzzy matching will benefit the same as normal matching. I guess the question is:

if an input fuzzy-matches `asdfg`, does it necessarily fuzzy-match `asdf`?

Thinking about it, this seems true; if something contains somewhere the characters asdfg in order, it also contains somewhere the characters asdf in order. 

The todo about case insensitivity is similar, but I think it probably doesn't matter as much (user is not likely to switch their input immediately from `something` to `somethinG`, which should be a refinement but will not be considered as one by the code).

DaveDavenport:
I think we can close this for now, as the other patches and work has made rofi fast enough :D thx.

PR #241, opened 2015-10-01
Fast ascii filtering

larkery:
I've done something again around the collation speed thing. Changes are:

 - in `menu()`, every input string is scanned for non-ascii codepoints.
 - this is to populate an array `lines_not_ascii` in `MenuState`
 - matching functions are supplied with an extra `not_ascii` parameter to tell them whether a particular line of input needs collating and case-folding or not

I haven't made it work in flex yet, because the flex rule will need rewriting a little (there is no strcasechr in the standard library to use in case-insensitive ascii mode).

Compared to the pre-collation approach I did before this:
 - doesn't store an extra copy of the input in memory; instead we have 1 int extra per input
 - doesn't add any noticeable startup time (the scan for non-ascii lines takes about a microsecond for half a million lines)

When most of the input is ascii it reduces the filtering time by about a thousandfold.

I've also very marginally improved the dmenu read speed by reducing the number of strlens and mallocs it does a little, so that you get to the menu slightly quicker.

### DIFF
--- a/source/dialogs/window.c
+++ b/source/dialogs/window.c
@@ -322,6 +322,7 @@ typedef struct _SwitcherModePrivateData
 } SwitcherModePrivateData;
 
 static int window_match ( char **tokens, __attribute__( ( unused ) ) const char *input,
+                          __attribute__( ( unused) ) int not_ascii,
                           int case_sensitive, unsigned int index, Switcher *sw )
 {
     SwitcherModePrivateData *rmpd = (SwitcherModePrivateData *) sw->private_data;
@@ -338,19 +339,19 @@ static int window_match ( char **tokens, __attribute__( ( unused ) ) const char 
             // e.g. when searching 'title element' and 'class element'
             char *ftokens[2] = { tokens[j], NULL };
             if ( !test && c->title[0] != '\0' ) {
-                test = token_match ( ftokens, c->title, case_sensitive, 0, NULL );
+                test = token_match ( ftokens, c->title, is_not_ascii(c->title), case_sensitive, 0, NULL );
             }
 
             if ( !test && c->class[0] != '\0' ) {
-                test = token_match ( ftokens, c->class, case_sensitive, 0, NULL );
+                test = token_match ( ftokens, c->class, is_not_ascii(c->title), case_sensitive, 0, NULL );
             }
 
             if ( !test && c->role[0] != '\0' ) {
-                test = token_match ( ftokens, c->role, case_sensitive, 0, NULL );
+                test = token_match ( ftokens, c->role, is_not_ascii(c->title), case_sensitive, 0, NULL );
             }
 
             if ( !test && c->name[0] != '\0' ) {
-                test = token_match ( ftokens, c->name, case_sensitive, 0, NULL );
+                test = token_match ( ftokens, c->name, is_not_ascii(c->title), case_sensitive, 0, NULL );
             }
 
             if ( test == 0 ) {
--- a/source/dialogs/window.c
+++ b/source/dialogs/window.c
@@ -343,15 +343,15 @@ static int window_match ( char **tokens, __attribute__( ( unused ) ) const char 
             }
 
             if ( !test && c->class[0] != '\0' ) {
-                test = token_match ( ftokens, c->class, is_not_ascii(c->title), case_sensitive, 0, NULL );
+                test = token_match ( ftokens, c->class, is_not_ascii(c->class), case_sensitive, 0, NULL );
             }
 
             if ( !test && c->role[0] != '\0' ) {
-                test = token_match ( ftokens, c->role, is_not_ascii(c->title), case_sensitive, 0, NULL );
+                test = token_match ( ftokens, c->role, is_not_ascii(c->role), case_sensitive, 0, NULL );
             }
 
             if ( !test && c->name[0] != '\0' ) {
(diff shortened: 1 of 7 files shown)


larkery:
I've fixed the fuzzy part as well - this now fuzzes correctly for unicode chars which it didn't before (it would match some pathological situations where a wide character with a couple of bytes like `b1 b2` is the token and the input line is some other wide characters like `b1 b3` ... `b4 b2`.

DaveDavenport:
nice!!

Hopefully I have some time this evening to play with it.

source/dialogs/window.c
```diff
@@ -338,19 +339,19 @@ static int window_match ( char **tokens, __attribute__( ( unused ) ) const char
             // e.g. when searching 'title element' and 'class element'
             char *ftokens[2] = { tokens[j], NULL };
             if ( !test && c->title[0] != '\0' ) {
-                test = token_match ( ftokens, c->title, case_sensitive, 0, NULL );
+                test = token_match ( ftokens, c->title, is_not_ascii(c->title), case_sensitive, 0, NULL );
             }
 
             if ( !test && c->class[0] != '\0' ) {
-                test = token_match ( ftokens, c->class, case_sensitive, 0, NULL );
+                test = token_match ( ftokens, c->class, is_not_ascii(c->title), case_sensitive, 0, NULL );
```
DaveDavenport: not ? is_not_ascii(c->class)

source/helper.c
```diff
@@ -306,53 +307,105 @@ int find_arg_char ( const char * const key, char *val )
...
+/*
+ * auxiliary to `fuzzy-token-match' below;
+ */
+static void advance_unicode_glyph( char** token_in, char** input_in ) {
+  // determine the end of the glyph from token
+
+  char *token = *token_in;
+  char *input = *input_in;
+
+  while (*token < 0) {
```
DaveDavenport: Do we target any platform where char is not signed? (given C does not specify)

source/dialogs/window.c
```diff
@@ -338,19 +339,19 @@ static int window_match ( char **tokens, __attribute__( ( unused ) ) const char
...
             }
 
             if ( !test && c->role[0] != '\0' ) {
-                test = token_match ( ftokens, c->role, case_sensitive, 0, NULL );
+                test = token_match ( ftokens, c->role, is_not_ascii(c->title), case_sensitive, 0, NULL );
```
DaveDavenport: is_not_ascii(c->role)?

source/dialogs/window.c
```diff
@@ -338,19 +339,19 @@ static int window_match ( char **tokens, __attribute__( ( unused ) ) const char
...
             }
 
             if ( !test && c->name[0] != '\0' ) {
-                test = token_match ( ftokens, c->name, case_sensitive, 0, NULL );
+                test = token_match ( ftokens, c->name, is_not_ascii(c->title), case_sensitive, 0, NULL );
```
DaveDavenport: is_not_ascii(c->name)

larkery:
As ever copy-and-paste causes the introduction of silly errors. Will repair directly. I guess for the signed-ness could cast to unsigned to force the issue and compare !=0 && < 12

DaveDavenport:
Not 100% sure? does ``` if(ch&0x80) {}```  work?  (check left most bit) (or cast to signed/unsigned char normally that is fairly inexpensive).

DaveDavenport:
Hmm did some test, on stock I can ready in 61 mbyte (1309056 lines) (count of monte cristo 24 times) it takes 0.66 seconds here on stock.  0.63 on this branch.

That is better then expected.

larkery:
The dmenu input will be fractionally quicker; others will be fractionally slower. The extra cost of scanning for ascii is offset in dmenu by not doing strlen a few times when reading the input.

I could pop the ascii scan into the first filter pass which would probably make startup time a few more microseconds quicker? Getting fast startup is really nice so it doesn't eat keypresses - additionally, could grab the keyboard before reading input? I think then perhaps even if it does take a half second to pop up you can start typing whilst it's reading. I'll do a quick test anyway.

The UTF8ness-test for a byte should probably be a macro in the helper.h as it's in two places. I'll do that.

I don't think the dmenu reading code can be made much quicker without doing something complicated like loading the lines asynchronously, which would be quite a big change.

larkery:
Ah, I see that grabbing the keyboard early is harder than I thought as well :/

larkery:
I tell a lie, it is easy after all, so long as not `-only-match` or `-no-custom` or `-select`

DaveDavenport:
I hope the last few commits/patches do not break to much.

people seems to be massifly confused by levenshtein sort, so might just remove it.

carnager:
@larkery since there is no Issues link on your fork, let me report it here:

```
menu=$(echo "1" | rofi -dmenu); echo "$menu"
```
this gives no output, which basically breaks around every dmenu script :smiling_imp:

larkery:
I've merged upstream, and I've fixed the radically bad mistake with dmenu; that was a side-effect of trying to grab keyboard early which seemed to work. Appears to work again now.

DaveDavenport:
@carnager is testing it for a bit..  If he gives it his approval I will merge it. It is a very nice improvement in  speed.

larkery:
OK - it doesn't have the other branch feature for faster re-filtering when the tokens to filter are a superset of the last query, but for me I think the simpler ascii input speed-up is enough anyway.
-----------------------

DaveDavenport:
subset is interesting too.

PR #253, opened 2015-10-29
Update README.md

keidax:
Add a reminder to reload your .Xresources file
(Helps with https://github.com/DaveDavenport/rofi/issues/110, which tripped me up)


DaveDavenport:
thanks.

PR #257, opened 2015-10-31
Add -no-parse-known-hosts option

klemens:
This option disables parsing of ~/.ssh/known_hosts.
For humans who like to keep the list of available servers clean. Also useful if you use aliases for the servers in ~/.ssh/config.

The previous default of always parsing known_hosts is not changed.


DaveDavenport:
thanks.

PR #271, opened 2015-11-18
Expand and resolve paths in script arguments

jasperla:
Currently an invocation such as:
`rofi -modi "window,run,ssh,Workspaces:~/.i3/i3_switch_workspaces.sh"`

would result in trying to execute the literal '~/.i3/<etc>' script, which it wouldn't find since the tilde wasn't expanded.
Now it will expand the tilde and resolve any relative paths.

### DIFF
--- a/source/dialogs/script.c
+++ b/source/dialogs/script.c
@@ -38,6 +38,8 @@
 #include "dialogs/script.h"
 #include "helper.h"
 
+static char *expand_script_path ( const char * );
+
 static char **get_script_output ( const char *command, unsigned int *length )
 {
     char **retv = NULL;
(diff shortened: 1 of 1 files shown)


DaveDavenport:
seems like `wordexp` is what we want.

jasperla:
Except for the fact that OpenBSD doesn't have it, and it's unlikely to ever get it. The way it's implemented in many places using a shellout is not going to be accepted into our libc. Having said that, nobody did a proper implementation either...

DaveDavenport:
aah that is a shame. it seems almost like we have to cook up something ourself.

DaveDavenport:
This might work:
```c
char * expand ( char *expand )                           
{                                                        
    char **str = g_strsplit(expand,G_DIR_SEPARATOR_S,-1);
    for (unsigned int i=0; str && str[i];i++){           
        // Replace ~ with current user homedir.          
        if(str[i][0] == '~' && str[i][1] == '\0'){       
            g_free(str[i]);                              
            str[i] = g_strdup(g_get_home_dir());         
        }                                                
        // If other user, ask getpwnam.                  
        else if ( str[i][0] == '~') {                    
            struct passwd *p = getpwnam(&(str[i][1]));   
            if(p != NULL){                               
                g_free(str[i]);                          
                str[i] = g_strdup(p->pw_dir);            
            }                                            
        }                                                
    }                                                    
    char *retv = g_build_filenamev(str);                 
    g_strfreev(str);                                     
    return retv;                                         
}                                                        
```

DaveDavenport:
Added to git.. lets see if this works.

jasperla:
Thanks, confirmed it works as intended!

PR #280, opened 2015-11-22
textbox: advance cursor correctly for multi-char input.

eigengrau:
Xutf8LookupString can return strings of length > 1, e. g. when an
XCompose sequence expands to more than one character. Accordingly,
we now check the length when advancing the cursor inside of the
textbox.


DaveDavenport:
Changed it a bit, to pass the length (return value Xutf8LookupString along the line instead of re-calculating).

Thanks.

Issue #183, opened 2015-07-08
rofi -show ssh doesn't work with termite


smlb:
As title say, when I try to launch ssh (with rofi.terminal: termite in ~/.Xresources), the terminal doesn't open. With urxvt it works like a charme. Any idea why this happens?

DaveDavenport:
I do not know? maybe termite needs a different command line?
Check the relevant configuration options:

```
rofi.ssh-client:                     ssh
rofi.ssh-command:                    {terminal} -e {ssh-client} {host}
```

smlb:
```
rofi.modi:                 run,window,ssh
!rofi.opacity:              75
rofi.width:                30
rofi.lines:                15
rofi.columns:              1
rofi.font:                 Terminus 9
rofi.fg:                   #839496
rofi.bg:                   #121212
rofi.bgalt:                #121212
rofi.hlfg:                 #5F87AF
rofi.hlbg:                 #242424
rofi.bc:                   #5B5B5B
rofi.terminal:             termite
rofi.ssh-client:           ssh
rofi.ssh-command:          {terminal} -e {ssh-client} {host}
rofi.run-command:          {cmd}
rofi.run-shell-command:    {terminal} -e {cmd}
rofi.disable-history:      false
rofi.levenshtein-sort:     false
rofi.case-sensitive:       false
rofi.auto-select:          false
rofi.parse-hosts:          false
```

This is how it looks like in my ~/.Xresources

DaveDavenport:
**you** need to update the ssh-command to fit with termite.

analogbyte:
For future visitors, quotes are needed with termite:

    rofi.ssh-command:          {terminal} -e "{ssh-client} {host}"

As this does not break rofi ssh for other terminal emulators, this might be a sane default option too.

Marcel-Robitaille:
I think @analogbyte 's solution should be default.

PR #284, opened 2015-11-23
config: quote terminal command-line when invoking ssh.

eigengrau:
At least termite will disregard {host} unless it’s quoted
as part of the argument to `-e`, and will immediately
exit after calling ssh without any arguments.

### DIFF
--- a/config/config.def.c
+++ b/config/config.def.c
@@ -81,7 +81,7 @@ Settings config = {
     .terminal_emulator = "rofi-sensible-terminal",
     .ssh_client        = "ssh",
     /** Command when executing ssh. */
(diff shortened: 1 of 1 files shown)


eigengrau:
Does this look sensible?

eigengrau:
Oh, this might be needed for the regular run command too.

DaveDavenport:
Please see #183 and #208 . 
You can fix  it yourself in your Xresources file or as cmd line flag.
Debian I know actually ships wrapper to 'fix' terminals.

eigengrau:
Aha, thanks! Wouldn’t have expected that it breaks rxvt.

DaveDavenport:
yeah it been annoying the hell out of me, why can't this be standarized by now.

PR #312, opened 2016-01-03
Fix a crash due to overlapping arguments of memcpy(3).

jasperla:
As the use of memcpy(3) on overlapping memory regions is strictly undefined,
OpenBSD's memcpy(3) triggers an abort in such cases.

### DIFF
--- a/source/rofi.c
+++ b/source/rofi.c
@@ -940,7 +940,7 @@ static void menu_refilter ( MenuState *state )
         g_mutex_clear ( &mutex );
         for ( unsigned int i = 0; i < nt; i++ ) {
             if ( j != states[i].start ) {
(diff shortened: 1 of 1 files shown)


DaveDavenport:
the source and destination region should not overlap (but yes it is within one allocated memory.

jasperla:
I was able to trigger it by starting rofi and having it generate the list of programs matching `e`, which is probably a rather common letter in program names :)

DaveDavenport:
Thanks for catching this, indeed a very silly mistake (again) on my part.

jasperla:
Thank you for the quick merge (again).

DaveDavenport:
almost done fixing all the Xfvb based tests :smile:

PR #321, opened 2016-01-17
added optional history; normalized variable names

DanteFireX:
I added the option to enable history in file browser mode. I changed the starting directory to a fixed directory. I found the old behavior not very intuitiv especially with enabled history. The old behavior can be restored easily. Additionally, I changed the variable names. Primarily, because the EDITOR variable could lead to a recursive call to EDITOR if xdg-open used this variable. Thus, I prefixed all the variables to create a new namespace for rofi. I am at most decent in bash scripting so don't expect to much, but it works for me and I thought it might be something others appreciate. I  hope it is of any use. :)

### DIFF
--- a/Examples/rofi-file-browser.sh
+++ b/Examples/rofi-file-browser.sh
@@ -1,46 +1,98 @@
 #!/usr/bin/env bash
 
-EDITOR=xdg-open
+# Various options for the file browser script:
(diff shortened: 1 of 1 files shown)


DaveDavenport:
Thanks.!

PR #330, opened 2016-02-02
Use correct namelen when connecting to i3 socket

t6:
This mirrors what i3 itself does and fixes i3 support on FreeBSD.

### DIFF
--- a/source/i3-support.c
+++ b/source/i3-support.c
@@ -50,7 +50,7 @@ char *i3_socket_path = NULL;
 void i3_support_focus_window ( Window id )
 {
     i3_ipc_header_t    head;
-    int                s, len;
+    int                s;
     ssize_t            t;
     struct sockaddr_un remote;
     size_t             upm = sizeof ( remote.sun_path );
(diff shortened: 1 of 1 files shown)


DaveDavenport:
I copied this from a working example. I changed ti for a reason, I cannot remember why though, Ill try to remember.

t6:
I've seen it in Beej's Guide like this: http://beej.us/guide/bgipc/output/html/multipage/unixsock.html#sockclient

But the man pages for connect(2) on FreeBSD and [OpenBSD](http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man2/connect.2?query=connect) imply that this is wrong.

Here is how i3 does it:
```c
    struct sockaddr_un addr;
    memset(&addr, 0, sizeof(struct sockaddr_un));
    addr.sun_family = AF_LOCAL;
    strncpy(addr.sun_path, socket_path, sizeof(addr.sun_path) - 1);
    if (connect(sockfd, (const struct sockaddr *)&addr, sizeof(struct sockaddr_un)) < 0)
        err(EXIT_FAILURE, "Could not connect to i3");
```

DaveDavenport:
Thanks for the info.  It seems to work fine  (nice travis is broken).

DaveDavenport:
Thanks for the patch.

PR #332, opened 2016-02-02
Use /usr/bin/env bash for rofi-sensible-terminal

t6:


DaveDavenport:
Thanks. (was copy pasted from i3-sensible-terminal).

Pyrohh:
AFAICT there are no bash constructs in that script; I just tested the following patch and it works fine:
```diff
diff --git a/script/rofi-sensible-terminal b/script/rofi-sensible-terminal
index c88e0c5..d0779c7 100755
--- a/script/rofi-sensible-terminal
+++ b/script/rofi-sensible-terminal
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 #
 # This code is released in public domain by Han Boetes <han@mijncomputer.nl>
 # Updated by Dave Davenport <qball@gmpclient.org>
```

t6:
You're right. `i3-sensible-terminal` is using `/bin/sh` on FreeBSD/OpenBSD too. I checked with Debian's checkbashisms script and it output:
```
possible bashism in /usr/local/bin/rofi-sensible-terminal line 13 ('command' with option other than -p):
    if command -v $terminal > /dev/null 2>&1; then
```
but I don't think this is a problem on the BSDs or Linux.

Pyrohh:
> but I don't think this is a problem on the BSDs or Linux.

Indeed, I forgot to mention I tested on OpenBSD. I'd guess it's only a problem with `dash`, assuming it truly is fully compliant on all possible fronts.

Issue #350, opened 2016-03-05
configure error


region51:
last git snapshot d59b9595b17c0d0c24fa2bd0370f1f75b606fe25
```
./configure: line 4891: syntax error near unexpected token `xcb-aux'
./configure: line 4891: `GW_CHECK_XCB(xcb-aux xcb-xkb xkbcommon xkbcommon-x11 xcb-ewmh xcb-xinerama xcb-icccm)'
```
depending on 9018d5d14fc82610b5329b28842858f762ae9e04
???

DaveDavenport:
Read the INSTALL.

```
git submodule update --init
```

```
autoreconf -i
```

region51:
@DaveDavenport :+1: 
` checking for GW_XCB_INTERNAL... yes`

DaveDavenport:
It is a very very large change. thanks for testing.

PR #365, opened 2016-03-26
Update `gh-pages` installation steps

benizi:
Ran into issue #350, because I was following the installation steps on
github.io, rather than locally.


DaveDavenport:
**NO!** this will make the website invalid for stable release. People that install git should be able to realize they should take instructions from git.
I will update this page **after** a release.

PR #367, opened 2016-03-27
Add staleness note to INSTALL.md git instructions

benizi:
Followup to #365 (and #350).  I didn't realize the `gh-pages` version of the instructions tracked the stable release.  It would be nice to know that the "Install a checkout from git" section on the site may be out of date.

Preview available at [the GitHub Pages version on my fork](https://benizi.github.io/rofi/p08-INSTALL.html#install-a-checkout-from-git) (only updated w/ this change, not the entire INSTALL.md).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -62,6 +62,11 @@ make install
 
 ## Install a checkout from git
 
+The GitHub Pages version of these directions may be out of date.  Please use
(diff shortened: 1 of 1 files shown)


DaveDavenport:
Thanks, this is perfect.

PR #374, opened 2016-04-05
rofi: Only display GIT_VERSION if available

sardemff7:
@DaveDavenport Is that ok for you?


sardemff7:
Pushed as 29f6823cf9bc482c1ceb04a08c5a71939e123197.

DaveDavenport:
thanks!

PR #381, opened 2016-04-10
Invalid UTF-8 in window titles crashes switcher. (Fix included)

blackhole89:
It appears that my music player echoes invalid UTF-8 from song titles in its title bar uncritically. This causes rofi's window switching function to segfault in strstr, as the invocation of `g_utf8_normalize` in `token_collate_key` returns NULL in that case and `normal_token_match` goes on to compare against it, apparently assuming it to be a valid pointer.

I've included one possible fix that did the trick for me, but you may prefer to fall back in some other way, try to strip bad UTF-8 altogether or validate the output of `token_collate_key` in all instances.

### DIFF
--- a/source/helper.c
+++ b/source/helper.c
@@ -148,6 +148,15 @@ char *token_collate_key ( const char *token, int case_sensitive )
     }
 
     compk = g_utf8_normalize ( tmp, -1, G_NORMALIZE_ALL );
+
+    // g_utf8_normalize may have failed; in that case, return the unnormalized string
+    if ( compk == NULL ) {
+        compk = tmp;
+    }
+    else {
+        g_free ( tmp );
+    }
+
     g_free ( tmp );
 
     return compk;
--- a/source/helper.c
+++ b/source/helper.c
@@ -157,8 +157,6 @@ char *token_collate_key ( const char *token, int case_sensitive )
         g_free ( tmp );
     }
 
-    g_free ( tmp );
-
     return compk;
 }
 void tokenize_free ( char ** tokens )


DaveDavenport:
This fix is not at the right place. I am actually amazed it fixed things, the font rendering and everything expects valid utf8. This should be fixed at the point it reads the value from the X server.

Will look at this later.

blackhole89:
Ah, correction - the UTF-8 in the song title and accordingly the window title is actually kosher, and what actually appears to be the cause of the problem is that it gets cut off after the 99th byte with no regard to UTF-8 codewords. When this slices through a codeword, invalid UTF-8 is generated. The offending line in 0.15.12 is `snprintf ( c->title, CLIENTTITLE, "%s", name );`.

It looks like HEAD has actually switched to dynamically allocated `c->title`, so the problem may be gone. Let me check - sorry for the spurious report if it turns out to be gone.

blackhole89:
Yep, can confirm it works now. Assuming `window_get_text_prop` is guaranteed to return valid UTF-8, the bug should be gone. Sorry I didn't check beforehand; the next most useful thing I can say is that `#define CLIENTTITLE         100` in that file no longer seems to serve any purpose.

DaveDavenport:
Brute-force way to fix things:
```patch
index eae59ce..a5e378f 100644
--- a/include/helper.h
+++ b/include/helper.h
@@ -166,5 +166,9 @@ void cmd_set_arguments ( int argc, char **argv );
 char *rofi_expand_path ( const char *input );
 unsigned int levenshtein ( const char *needle, const char *haystack );
 
+/**
+ * Convert string to valid utf-8, replacing invalid parts with replacement character.
+ */
+char * rofi_force_utf8 ( gchar *data );
 /*@}*/
 #endif // ROFI_HELPER_H
diff --git a/source/dialogs/window.c b/source/dialogs/window.c
index 5c2bdc1..0c8b142 100644
--- a/source/dialogs/window.c
+++ b/source/dialogs/window.c
@@ -295,6 +295,14 @@ static client* window_client ( xcb_window_t win )
         c->hint_flags = r.flags;
     }
 
+    /** Do UTF-8 Check */
+    {
+        c->title = rofi_force_utf8 ( c->title );
+        c->class = rofi_force_utf8 ( c->class );
+        c->name  = rofi_force_utf8 ( c->name );
+        c->role  = rofi_force_utf8 ( c->role );
+    }
+
     winlist_append ( cache_client, c->window, c );
     g_free ( attr );
     return c;
index 2485e80..8aee921 100644
--- a/source/helper.c
+++ b/source/helper.c
@@ -659,3 +659,38 @@ unsigned int levenshtein ( const char *needle, const char *haystack )
     }
     return column[needlelen];
 }
+
+char * rofi_force_utf8 ( gchar *start )
+{
+    if ( start == NULL ) {
+        return NULL;
+    }
+    const char        *data = start;
+    const char *end;
+    gsize       length = strlen(data);
+    GString     *string;
+
+    if ( g_utf8_validate ( data, length, &end ) ) {
+        return start;
+    }
+printf("invalid\n");
+    string = g_string_sized_new ( length + 16 );
+    do {
+        /* Valid part of the string */
+        g_string_append_len ( string, data, end - data );
+
+        /* Replacement character */
+        g_string_append ( string, "�" );
+
+        length -= ( end - data ) + 1;
+        data    = end + 1;
+    } while ( !g_utf8_validate ( data, length, &end ) );
+
+    if ( length ) {
+        g_string_append_len ( string, data, length );
+    }
+
+    // Free input string.
+    g_free ( start );
+    return g_string_free ( string, FALSE );
+}
```

PR #384, opened 2016-04-23
Permit binding a modifier key on its own

gbitzes:
Use case: I have bound Super_L to launch rofi, and I want that the same button also closes the window. Specifying "-kb-cancel Escape,Super_L" does not currently work, because x11_parse_key does not support binding to a single modifier key.

### DIFF
--- a/source/x11-helper.c
+++ b/source/x11-helper.c
@@ -438,14 +438,22 @@ gboolean x11_parse_key ( char *combo, unsigned int *mod, xkb_keysym_t *key )
         seen_mod = TRUE;
     }
 
-    *mod = modmask;
-
-    // Skip modifier (if exist) and parse key.
+    // Find location of modifier (if it exists)
     char i = strlen ( combo );
 
     while ( i > 0 && !strchr ( "-+", combo[i - 1] ) ) {
         i--;
     }
+
+    // if there's no "-" or "+", we might be binding directly to a modifier key - no modmask
+    if( i == 0 ) {
+        *mod = 0;
+    }
+    else {
+        *mod = modmask;
+    }
+
+    // Parse key
     xkb_keysym_t sym = XKB_KEY_NoSymbol;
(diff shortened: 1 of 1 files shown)


DaveDavenport:
Looks useful, thinking what could go wrong by this. (given keyboards, layouts and bindings can be tricky to get right).
Only one I can think of for now is (for the user) unexpected behavior if they are foolish enough to bind both Super_L and Super+Something. We could probably add a sanity check to this.

gbitzes:
I agree, thinking about keymaps and modmasks always makes my head hurt.. I'm not sure if it's possible to have a robust sanity check. 

Checking for conflicts from binding eg both Super_L and super+a would involve testing the output of x11_parse_key against all previous assignments:

```
   for every new key-binding (key1, mask1)
        for every previous key-binding (key2, mask2)
            check for conflict between (key1, mask1), (key2, mask2)
```

You could check for conflicts by translating key1, key2 to their own modmasks if any, (mod3, mod4), and doing some bitwise tricks with mask1 and mask2.

The translation part is not as straight-forward as it may seem. A user could assign the keysym "m" to act as Mod1, for example, so testing for "Alt_L" and "Alt_R" is not enough, we need to know the modifier assignments of the current configuration. Not sure if possible or worth doing..

sardemff7:
If we merge the bind-on-release branch, we can make these a special case. Modifiers would always bind-on-release, but not trigger if something else was pressed in-between. (For normal bindings, bind-on-release trigger all actions that was pressed.)

DaveDavenport:
we should probably do the merge of the bind-on-release branch after this weekends bug fix release and make sure this functionality is in there.

sardemff7:
I rebased locally, on top on the binding-on-release, and it works perfectly.

Pushed as 8af7f014cdc06f6b6ac9d6c6796a143e6307e7dd.

DaveDavenport:
thx.

PR #393, opened 2016-04-28
When switching windows also switch the desktop

domidimi:
If the window we switch to is not on the same desktop, first switch to
the desktop.

### DIFF
--- a/source/dialogs/window.c
+++ b/source/dialogs/window.c
@@ -37,6 +37,7 @@
 #include <xcb/xcb.h>
 #include <xcb/xcb_ewmh.h>
 #include <xcb/xcb_icccm.h>
+#include <xcb/xcb_atom.h>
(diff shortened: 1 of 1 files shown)


DaveDavenport:
Is this still needed? for me it works without in I3?

domidimi:
This is in the else part. I'm using awesome not i3.

DaveDavenport:
In latest i3,  you do not need i3 support anymore as it obeys EWMH signals.

I will try to check a few Window managers later, but I suspect the patch will not result into any issues.

Thanks!

domidimi:
I got your point. I don't know if the behavior of awesome is correct. I noticed now that the screen is also not getting switched in awesome. I have to see if that functionality is in the C or Lua code. It'll take some time.

PR #412, opened 2016-05-30
Added the descriptions of the key bindings

mardukbp:
I reformatted the entries for the key bindings so that I could have the vim window with the rofi source and the window with the manual open side by side.

I could not understand from the code what other uses does the Tab key have.

I think it would be nice to use a [macro](https://github.com/eudoxia0/cmacro) to generate the code for the custom keybindings.

I think the current output format is clearer than the previous one.

Feel free to modify anything.

### DIFF
--- a/source/keyb.c
+++ b/source/keyb.c
@@ -34,57 +34,159 @@ ActionBindingEntry abe[NUM_ABE];
  */
 DefaultBinding bindings[NUM_ABE] =
 {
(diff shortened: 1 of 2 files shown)


mardukbp:
I just noticed in README.md that some key bindings have different actions depending on the modi or if rofi is run with -dmenu. Could we have a height of three rows per entry?

DaveDavenport:
thanks, merged the descriptions.

PR #445, opened 2016-08-14
Use bash instead of sh for get_git_rev.sh

haasn:
The script relies on bash extensions to `echo`, which is not portable to
POSIX shells. Simplest fix is to request bash explicitly.


haasn:
Note: In particular, this breaks when /bin/sh is dash, because `-n` is treated as a regular word.

DaveDavenport:
Thanks.

PR #452, opened 2016-08-25
Bind Control+g to kb-cancel

fice-t:
Control+g is the readline standard for abort, so it would be nice
to have this by default.


DaveDavenport:
thanks.

PR #455, opened 2016-08-31
add xinerama support

seletskiy:
My Xorg setup doesn't support RANDR (nvidia+xorg), so I basically get error
when trying to run `rofi`. It's because `rofi` requests `randr` info and
break xcb connection by doing that.

```
xkbcommon: ERROR: xkb_x11_keymap_new_from_device: illegal device ID: -1
Failed to get Keymap for current keyboard device.
*** Error in `rofi': double free or corruption (top): 0x000000000098f7c0 ***
```

### DIFF
--- a/include/x11-helper.h
+++ b/include/x11-helper.h
@@ -153,6 +153,8 @@ cairo_surface_t * x11_helper_get_bg_surface ( void );
  * Creates an internal represenation of the available monitors.
  * Used for positioning rofi.
  */
+uint8_t x11_is_randr_present ( void );
+void x11_build_monitor_layout_xinerama ( void );
 void x11_build_monitor_layout ( void );
 void x11_dump_monitor_layout ( void );
 /*@}*/
--- a/include/x11-helper.h
+++ b/include/x11-helper.h
@@ -153,8 +153,6 @@ cairo_surface_t * x11_helper_get_bg_surface ( void );
  * Creates an internal represenation of the available monitors.
  * Used for positioning rofi.
  */
-uint8_t x11_is_randr_present ( void );
-void x11_build_monitor_layout_xinerama ( void );
 void x11_build_monitor_layout ( void );
 void x11_dump_monitor_layout ( void );
 /*@}*/
--- a/source/x11-helper.c
+++ b/source/x11-helper.c
@@ -37,7 +37,9 @@
 
 #include <xcb/xcb.h>
 #include <xcb/randr.h>
+#include <xcb/xinerama.h>
 #include <xcb/xcb_ewmh.h>
+#include <xcb/xproto.h>
 #include "xcb-internal.h"
 #include "xcb.h"
 #include "settings.h"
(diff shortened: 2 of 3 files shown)


DaveDavenport:
Fun just removed xinerama and went for randr because it caused so much issue.
Will add some notes to patch.

source/x11-helper.c
```diff
@@ -213,8 +215,66 @@ static workarea * x11_get_monitor_from_output ( xcb_randr_output_t out )
...
 }
 
+uint8_t x11_is_randr_present () {
+    xcb_query_extension_cookie_t randr_cookie = xcb_query_extension (
+        xcb->connection,
+        sizeof("RANDR")-1,
+        "RANDR"
+    );
+
+    xcb_query_extension_reply_t *randr_reply = xcb_query_extension_reply (
```
DaveDavenport: randr_reply is never free'ed:
```
* The returned value must be freed by the caller using free().
```

source/x11-helper.c
```diff
@@ -213,8 +215,66 @@ static workarea * x11_get_monitor_from_output ( xcb_randr_output_t out )
     return retv;
 }
 
+uint8_t x11_is_randr_present () {
```
DaveDavenport: Please use `int` as return value, this is normally used in rofi for returning TRUE/FALSE;

include/x11-helper.h
```diff
@@ -153,6 +153,8 @@ cairo_surface_t * x11_helper_get_bg_surface ( void );
  * Creates an internal represenation of the available monitors.
  * Used for positioning rofi.
  */
+uint8_t x11_is_randr_present ( void );
+void x11_build_monitor_layout_xinerama ( void );
```
DaveDavenport: No need to export these functions outside of x11-helper.c.

DaveDavenport:
Thanks, this will be a useful addition.

source/x11-helper.c
```diff
@@ -213,8 +215,66 @@ static workarea * x11_get_monitor_from_output ( xcb_randr_output_t out )
...
+    }
+
+    free ( screens_reply );
+}
+
 void x11_build_monitor_layout ()
 {
+    if ( !x11_is_randr_present () ) {
+        fprintf ( stderr, "Using XINERAMA instead of XRANDR\n" );
+        x11_build_monitor_layout_xinerama ();
```
DaveDavenport: Might want to check if xinerama is available.

seletskiy: Yeah, but if RANDR is not present and XINERAMA is missing to we will be out of lack anyway.

seletskiy:
@DaveDavenport: Sorry for the mess, issues should be fixed right now.

source/x11-helper.c
DaveDavenport: No, if x11_build_monitor_layout, it falls back to the size of the root window.

DaveDavenport:
No problem, very happy with the contribution.

kovetskiy:
@seletskiy @DaveDavenport please add new dependency to ArchLinux PKGBUILD.

DaveDavenport:
I tweaked some things, hopefully did not break anything.

@kovetskiy that is not in my control.

seletskiy:
@DaveDavenport: Seems to be fine.

DaveDavenport:
Thanks again for the contribution.

PR #486, opened 2016-10-23
changed subprocess call in file-browser example

flhp:
Hi, this is a really tiny thing but it's been bothering for a while:
I'm using the file-browser script as a quick way to open files, but xdg-open always seemed to open the wrong programs (mostly firefox, but for everything).
After spending way too much time on this, I traced it to the "coproc" call. When used in a bash shell, it works normally, but called from a script it seems to mess with the way xdg-open opens files.
Anyways, removing the first ampersand from that line fixes it.
From the bash manpage: 
```
There are two formats for redirecting standard output and standard error:

              &>word
       and
              >&word

Of the two forms, the first is preferred.  This is semantically equivalent to

              >word 2>&1
```
So to me, it seems either the first ampersand or 2nd redirect is superfluous.

Just one character, but this has itched me for some time now and maybe it helps someone else.

### DIFF
--- a/Examples/rofi-file-browser.sh
+++ b/Examples/rofi-file-browser.sh
@@ -70,7 +70,7 @@ then
             fi
         fi
         # Open the selected entry with $ROFI_FB_GENERIC_FO
-        coproc ( "${ROFI_FB_GENERIC_FO}" "${ROFI_FB_CUR_DIR}" & > /dev/null  2>&1 )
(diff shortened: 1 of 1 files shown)


DaveDavenport:
I agree the & can be removed, it looks like it is an old thing left over.

I don't think it was original mend to be part of the redirect.. (see the extra spaces!)
it just screwed up things.

Also, why just fix one occurrence? will merge when second also fixed.

DaveDavenport:
fixed it. Thanks.

PR #525, opened 2017-01-04
Input location should be configurable

milouse:
One may want to keep input bar above results, even if rofi start at the bottom of the screen, or in the contrary input bar bellow results when rofi run at the top of the screen.

### DIFF
--- a/config/config.c
+++ b/config/config.c
@@ -76,6 +76,14 @@ Settings config = {
      *
      */
     .location          = WL_CENTER,
+    /**
+     * Location of the input line.
+     * Enumeration indicating location of the input line
+     *
+     * IL_TOP       IL_BOTTOM
+     *
+     */
+    .input_location          = IL_TOP,
     /** Padding between elements */
     .padding           =                                         5,
     /** Y offset */
--- a/doc/rofi-manpage.markdown
+++ b/doc/rofi-manpage.markdown
@@ -13,6 +13,7 @@
 [ -font *pangofont* ]
 [ -terminal *terminal* ]
 [ -location *position* ]
+[ -input-location *position* ]
 [ -fixed-num-lines ]
 [ -padding *padding* ]
 [ -display *display* ]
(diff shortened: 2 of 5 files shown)


DaveDavenport:
This will be, at some point, be part of the new theme format (themer branch).

DaveDavenport:
Currently in theme branch you can set the each widget to be packed at begin or end.
That together with the reverse option on the listview allows you to swap this around.

Example will be provided when merged in main-branch.

milouse:
Ok thanks for this information, I didn't know.

However, a theme will give this option for all instances of rofi, not just one, given a command line switch ?

DaveDavenport:
Why would you want to switch this between running instances. That just sounds confusing.

DaveDavenport:
I don't think it is a good idea to make everything a command-line option. This does not work (given evidence from the past).

You can override the theme from the command-line, so advanced setups can still do this. The other 99% will be happy having rofi in one consistent place.

DaveDavenport:
Added a commandline option that allows to override part of the theme file.
For example to move the textbox to first position (depending on the theme):

`-theme-str "#window mainbox textbox { index: 0; }"`

DaveDavenport:
Last updates allow you to set it anywhere.
Like left of listview: https://filebin.pw/6c2QT/
or below prompt: https://filebin.pw/EiiQ/

cipriancraciun:
For some reason this doesn't work.

I've tried everything on the line:
~~~~
rofi -theme-str 'listview { index : 0; }' -show run
rofi -theme-str 'inputbar { index: 1; }' -show run
rofi -theme-str 'inputbar { index: 1; } listview { index : 0; }' -show run
~~~~

(If I use `background-color: red` it does change that widget, so my selector seems to be OK.)

sardemff7:
Please open a new issue if you have a new problem. You may want to check with people on IRC about theming first, and help improve the documentation.

DaveDavenport:
This options has been replaced years ago when the theme format became more solid. 
The interface can now be customized completely using horizontal/vertical boxes and packed children. See `rofi-theme` manpage.

PR #533, opened 2017-01-09
Revise fuzzy finding algorithm for -matching fuzzy

MaskRay:
This is my attempt to address https://github.com/DaveDavenport/rofi/issues/480 , it revises the `-matching fuzzy` algorithm by incorporating ideas from https://github.com/jhawthorn/fzy/blob/master/ALGORITHM.md and https://github.com/junegunn/fzf/blob/master/src/algo/algo.go .

I test the revised algorithm by matching paths with the following script `rofi-file-browser`:

```zsh
#!/bin/zsh
f=$2
prev=/tmp/.rofi-file-browser
n=5000

my_find() {
  find $1 -mindepth 1 -printf '%P\n' | head -n $n
}

if [[ -z $f ]]; then
  rm -f $prev
  my_find $1
else
  if [[ -f $prev ]]; then
    cd $(<$prev)
  else
    cd $1
  fi
  if [[ -d $f ]]; then
    cd $f
    print -r -- $PWD > $prev
    echo ..
    my_find
  else
    rm -f $prev
    coproc xdg-open $f
    disown
  fi
fi
```

`./rofi -show file -matching fuzzy -modi "file:rofi-file-browser $HOME"`

### DIFF
--- a/source/view.c
+++ b/source/view.c
@@ -25,6 +25,7 @@
  */
 
 #include <config.h>
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -62,6 +63,17 @@
 
 /** The Rofi View log domain */
 #define LOG_DOMAIN    "View"
+
+#define FUZZY_SCORER_MAX_LENGTH      256
+#define MIN_SCORE                    (INT_MIN / 2)
+#define LEADING_GAP_SCORE            -4
+#define GAP_SCORE                    -5
+#define WORD_START_SCORE             50
+#define NON_WORD_SCORE               40
+#define CAMEL_SCORE                  (WORD_START_SCORE + GAP_SCORE - 1)
+#define CONSECUTIVE_SCORE            (WORD_START_SCORE + GAP_SCORE)
+#define PATTERN_NON_START_MULTIPLIER 1
+#define PATTERN_START_MULTIPLIER     2
 
 #include "xcb.h"
 /**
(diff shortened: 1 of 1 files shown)


DaveDavenport:
Thanks for the patch, looks very interresting.

I saw you added it as a matching mode, while it is only doing sorting. Is it possible to add this as a sorting method (maybe also using for non-fuzzy matching methods?)?

source/view.c
```diff
@@ -540,25 +553,95 @@ static void rofi_view_call_thread ( gpointer data, gpointer user_data )
...
-    // input changed
+    char *pattern = mode_preprocess_input(t->state->sw, t->state->text->text);
+    glong plen = g_utf8_strlen(pattern, -1);
     for ( unsigned int i = t->start; i < t->stop; i++ ) {
         int match = mode_token_match ( t->state->sw, t->state->tokens, i );
         // If each token was matched, add it to list.
         if ( match ) {
             t->state->line_map[t->start + t->count] = i;
-            if ( config.levenshtein_sort ) {
+            if (config.matching_method == MM_FUZZY) {
```
DaveDavenport: This whole part should be placed in a subfunction.  It is rather a large inline chunk.

source/view.c
```diff
@@ -540,25 +553,95 @@ static void rofi_view_call_thread ( gpointer data, gpointer user_data )
...
+    return CAMEL_SCORE;
+  if (curr == NON_WORD)
+    return NON_WORD_SCORE;
+  return 0;
+}
+
 static void filter_elements ( thread_state *t, G_GNUC_UNUSED gpointer user_data )
 {
-    // input changed
+    char *pattern = mode_preprocess_input(t->state->sw, t->state->text->text);
```
DaveDavenport: Might want to reduce the scope of this.  If we don't do sorting, we don't want to call this.

source/view.c
```diff
@@ -972,11 +1055,18 @@ static void rofi_view_refilter ( RofiViewState *state )
         tokenize_free ( state->tokens );
         state->tokens = NULL;
     }
+    if ( state->pattern ) {
```
DaveDavenport: state->pattern is never used? it is only set and free'ed?

source/view.c
```diff
@@ -540,25 +553,95 @@ static void rofi_view_call_thread ( gpointer data, gpointer user_data )
     g_mutex_unlock ( t->mutex );
 }
 
+enum CharClass { LOWER, UPPER, DIGIT, NON_WORD };
+
+static enum CharClass char_class(gunichar c)
```
DaveDavenport: can this be named more clearly:  something like: `rofi_sort_fzf_get_character_class`. Same for `scrore_for`.

source/view.c
```diff
@@ -540,25 +553,95 @@ static void rofi_view_call_thread ( gpointer data, gpointer user_data )
...
+                    gchar *pit = pattern, *sit;
+                    enum CharClass prev = NON_WORD, cur;
+                    for (si = 0, sit = str; si < slen; si++, sit = g_utf8_next_char(sit)) {
+                        cur = char_class(g_utf8_get_char(sit));
+                        score[si] = score_for(prev, cur);
+                        prev = cur;
+                    }
+                    for (pi = 0; pi < plen; pi++, pit = g_utf8_next_char(pit)) {
+                        int prefix = MIN_SCORE, ul, ul2, uls, uls2;
+                        if (g_utf8_get_char(pit) == ' ')
```
DaveDavenport: Rofi coding style is to always add {} .
(there is a make indent target  you can run if you have uncrustify installed).

DaveDavenport: g_unichar_isspace better here?

source/view.c
MaskRay: `state->pattern` is the complete pattern (`state->tokens` is its tokenized form) which is only used in `-matching fuzzy` mode.

source/view.c
MaskRay: And I'm also not sure whether I should switch to the verbose space surrounded `if ( x )` style...

source/view.c
MaskRay: I'll change the name to `rofi_sort_get_character_class` because it is not a re-implementation of `fzf`'s fuzzy finding algorithm. (I have read its code and thought it difficult to migrate its code to rofi because it is too complex.)

source/view.c
DaveDavenport: I must be missing something, I don't see where it is used. I just see it being set and free'ed.

MaskRay: I find `state->tokens` hard to use in the revised fuzzy finding algorithm because it requires the complete input so I added `char* pattern` to `view-internal.h`

DaveDavenport: Yes, but in the patch it is not used. You redefine pattern in filter_elements, but state->filter is not used.

DaveDavenport:
Just wondering, if the patch complete? i would expect changed to helper.c.

source/view.c
MaskRay: OK, I've moved this part to `fzf_scorer_fuzzy_evaluate`.

DaveDavenport:
Patch is looking good. 
Would be nice if the sorting done could also be applied to the other matchers.
Do you think this is possible?

MaskRay:
Yes, this patch is complete for revising `-matching fuzzy`. The scorer applies when the pattern is a subsequence of the text.

DaveDavenport:
Ok. I think (_but I can do that later_) that I should make a global 'sort' option and the user can pick either levenshtein or this algorithm. (maybe only functional on fuzzy). 
So the keybinding for enable/disable sort still works.

There are some compile warnings about uninitialized variables and missing () around expression.

```
../source/view.c: In function 'rofi_scorer_get_score_for':
../source/view.c:572:21: warning: suggest parentheses around '&&' within '||' [-Wparentheses]
   if (prev == LOWER && curr == UPPER ||
       ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~
  CC       source/widgets/rofi-textbox.o
../source/view.c: In function 'filter_elements':
../source/view.c:614:61: warning: 'ulefts' may be used uninitialized in this function [-Wmaybe-uninitialized]
                     : MAX(uleft + CONSECUTIVE_SCORE, ulefts + t);
                                                             ^
../source/view.c:587:16: note: 'ulefts' was declared here
         uleft, ulefts, // uleft: value of the upper left cell; ulefts: maximum value of uleft and cells on the left
                ^~~~~~
In file included from /usr/lib/glib-2.0/include/glibconfig.h:9:0,
                 from /usr/include/glib-2.0/glib/gtypes.h:32,
                 from /usr/include/glib-2.0/glib/galloca.h:32,
                 from /usr/include/glib-2.0/glib.h:30,
                 from ../include/settings.h:4,
                 from ../source/view.c:48:
/usr/include/glib-2.0/glib/gmacros.h:288:26: warning: 'uleft' may be used uninitialized in this function [-Wmaybe-uninitialized]
 #define MAX(a, b)  (((a) > (b)) ? (a) : (b))
                          ^
../source/view.c:587:9: note: 'uleft' was declared here
         uleft, ulefts, // uleft: value of the upper left cell; ulefts: maximum value of uleft and cells on the left
         ^~~~~
../source/view.c:597:5: warning: 'plen' may be used uninitialized in this function [-Wmaybe-uninitialized]
     for (pi = 0; pi < plen; pi++, pit = g_utf8_next_char(pit)) {
     ^~~
../source/view.c:634:11: note: 'plen' was declared here
     glong plen;
```
Might be good to just 0 initialize them.

One other problem (but levenshtein has the same issue) is that the matcher might also match against non-visible data. But I guess this is not big issue.

I will play a bit with the patch (I am very curious how it behaves ) and I will merge it later.

Thanks for the great work!

source/view.c
```diff
@@ -540,25 +552,114 @@ static void rofi_view_call_thread ( gpointer data, gpointer user_data )
...
+    return lefts;
+}
+
 static void filter_elements ( thread_state *t, G_GNUC_UNUSED gpointer user_data )
 {
-    // input changed
+    char *pattern = NULL;
+    glong plen;
+    if (config.matching_method == MM_FUZZY) {
+        pattern = mode_preprocess_input(t->state->sw, t->state->text->text);
```
DaveDavenport: pattern is also needed for levenhtein sort. Now levenshtein sort is broken.

MaskRay: Updated. I've also added some comments to the algorithm.

MaskRay:
Thanks! The following three lines are left for debugging purpose, please remove them... 

```c
    if ( plen == 5 ) {
        plen = plen;
    }
```

DaveDavenport:
Kay I will. doing some cleanups anyway ( indenting, commenting etc.)

DaveDavenport:
Thanks for the patch.

PR #541, opened 2017-01-20
Usability enhancements

Calrama:
#### Bang mode prefixes
This is required to match on modes that share a prefix.
Let 'power' and 'pass' be such modes for the following explanation:
Previously, only the first character after the bang was compared,
so '!p' would always be matched to the 'pass' mode and there was no
way to limit selection in combi mode to the 'power' mode.
Now we can use prefixes of arbitrary length following the bang
such as '!po' (matches 'power' mode), or '!pa' (matches 'pass' mode).
Prefixes of length 1 are unchanged compared to the previous
behaviour, i.e. '!p' will still match 'pass'.

#### Automatic DPI scaling
If no specific dpi is requested match the one of the X11
screen we connected to by default.
This is necessary for out-of-the-box support of HIDPI monitors,
especially when the user - e.g. on a laptop with such
an internal monitor - frequently switches between internal
and external (normal dpi) monitors.

### DIFF
--- a/source/dialogs/combi.c
+++ b/source/dialogs/combi.c
@@ -237,14 +237,20 @@ static char * combi_preprocess_input ( Mode *sw, const char *input )
 {
     CombiModePrivateData *pd = mode_get_private_data ( sw );
     pd->current = NULL;
-    if ( input != NULL && input[0] == '!' && strlen ( input ) > 1 ) {
-        for ( unsigned i = 0; i < pd->num_switchers; i++ ) {
-            if ( input[1] == mode_get_name ( pd->switchers[i] )[0] ) {
-                pd->current = pd->switchers[i];
-                if ( input[2] == '\0' ) {
-                    return NULL;
+    if ( input != NULL && input[0] == '!' ) {
+        char *eob = strchrnul ( input, ' ' );
+        ssize_t bang_len = eob - input - 1;
+        if ( bang_len > 0 ) {
+            for ( unsigned i = 0; i < pd->num_switchers; i++ ) {
+                const char *mode_name = mode_get_name ( pd->switchers[i] );
+                size_t mode_name_len = strlen ( mode_name );
+                if ( (size_t) bang_len <= mode_name_len && strncmp ( &input[1], mode_name, bang_len ) == 0 ) {
+                    pd->current = pd->switchers[i];
+                    if ( eob[0] == '\0' || eob[1] == '\0' ) {
+                        return NULL;
+                    }
+                    return g_strdup ( eob+1 );
                 }
-                return g_strdup ( &input[2] );
             }
         }
     }
--- a/source/view.c
+++ b/source/view.c
@@ -667,9 +667,14 @@ void __create_window ( MenuFlags menu_flags )
     // Set the font options from the xlib surface
     pango_cairo_context_set_font_options ( p, fo );
     // Setup dpi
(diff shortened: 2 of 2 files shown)


DaveDavenport:
Please make a merge request for each individual  item. Meta PRs containing multiple unrelated changes will not be accepted.

1.) The bang patch is interesting and a welcome improvement. The patch is not ready, as it does the longer prefix while matching, but not when the final result is accepted  (`combi_mode_result`) or when computing the completion (`combi_get_completion`).
It might also be good to do this utf-8 aware.

2.) Very useful improvement. But it changes default behavior. I would prefer keeping current behavior and auto-size when dpi is configured negative.  (Changing the default behavior always gives a storm flood of users complaining.)
Might be nice to calculate the DPI not based on the xcb->screen but on the actual monitor we display on?

DaveDavenport:
Added DPI option to git. If dpi is set to 0, it tries to auto-detect it based on monitor placed on, and as fallback to screen.

Calrama:
1.) I can put this in a separate PR, of course, but I'll have to look into those functions again, since it works as expected for me. I'm also not certain what you mean by utf-8 aware. The utf-8 space U+0020 matches the ASCII one, same for null U+0000, so afaict the implementation should work fine for any utf-8 input.
2.) The issue with trying to detect the monitor's dpi instead of using the screen's one is that under X11 the screen's one is the correct one. AFAIK X11 does not support monitors (i.e. X Randr outputs) on the same screen with different dpi values, so using anything other than the screen's dpi is wrong. The screen's dpi is also what all other software I'm aware of that's not GDK-based (such as QT5, i3wm, ...) use, because it can easily be corrected via `xrandr --dpi` if, for example, the hardware reports wrong values. Please don't try to use the monitor's value and use the screen's one. That makes rofi behave as expected (i.e. it fits in nicely with all the other software).

DaveDavenport:
Don't have to much experience with DPI stuff. 
I just liked that on my 216dpi tablet screen the text was the same size as on the external (102dpi) monitor. 

I will see if I can make these two separate options.. one that mirrors gtk and others one that does it unconventional per-monitor.

Calrama:
Hm, I've researched a bit more and it seems QT5, at least in the newest version, also uses a per-monitor approach, so I take back the "incorrect" judgement, my apologies. That does leave one more problem, though: On smaller screens - i.e. those where the distance between you and the screen is relatively small - you don't necessarily want the "correct" scaling. For example, a 13" QHD+ laptop monitor has a dpi of ~277, but I use a `xrandr --dpi 177` instead, since I'm a lot closer to it than to a 96 dpi 24" external monitor, which means the subjective size of everything on the monitor is that much larger (=> perspective), so if everything were scaled to the correct pixel size, it would (and does) seem huge.
So I agree that they should be separate options, maybe like this (expands the second case of what you already implemented):
- `config.dpi >= 0` => use that value
- `config.dpi = 0 && config.use_monitor_dpi` => use monitor dpi value
- `config.dpi = 0 && !config.use_monitor_dpi` => use screen dpi value
- `config.dpi < 0` => don't scale
And `config.use_monitor_dpi` should probably default to `1`.

IvanMalison:
@Calrama @DaveDavenport -- What's the status of the dpi settings?

DaveDavenport:
Good. see manpage in git.

PR #542, opened 2017-01-21
Combi mode: Bang mode prefixes with len>1

Calrama:
This is required to match on modes that share a prefix.
Let 'power' and 'pass' be such modes for the following explanation:
Previously, only the first character of after the bang was compared,
so '!p' would always be matched to the 'pass' mode and there was no
way to limit selection in combi mode to the 'power' mode.
Now we can use prefixes of arbitrary length following the bang
such as '!po' (matches 'power' mode), or '!pa' (matches 'pass' mode).
Prefixes of length 1 are unchanged compared to the previous
behaviour, i.e. '!p' will still match 'pass'.

### DIFF
--- a/include/helper.h
+++ b/include/helper.h
@@ -221,4 +221,17 @@ PangoAttrList *token_match_get_pango_attr ( ThemeHighlight th, GRegex **tokens, 
  */
 int rofi_scorer_fuzzy_evaluate ( const char *pattern, glong plen, const char *str, glong slen );
 /*@}*/
+
+/**
+ * @param a    UTF-8 string to compare
+ * @param b    UTF-8 string to compare
+ * @param n    Maximum number of characters to compare
+ *
+ * Compares the `G_NORMALIZE_ALL_COMPOSE` forms of the two strings.
+ *
+ * @returns less than, equal to, or greater than zero if the first `n` characters (not bytes) of `a`
+ *          are found, respectively, to be less than, to match, or be greater than the first `n`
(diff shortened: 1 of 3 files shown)


Calrama:
One thing I've noticed is that rofi does not show completions for all matched modes in combi-mode immediately (and this PR has - AFAICT - no impact on that).
Example: If you're running rofi with, e.g. "drun" and "run" as the combi-modi then entering '!d' (after this change also '!dr', '!dru', and '!drun') will show all available completions for "drun" at once (this is as expected). If, however, one enters '!r' (after this change also '!ru', and '!run') nothing will get shown until one enters some non-space (non-bang prefix) input to match against.

DaveDavenport:
about the utf8, you want to match on character, not byte.
For example (sorry quick thrown together example):

```bash
./rofi -show "combi" -combi-modi "àt:~/Programming/GitHub/rofi/Examples/rofi-file-browser.sh,a:ls" 
```
If I type !a tit matches the `àt` browser.  If I type !at it does not match. 
You actually want `!a` to match the ls.

DaveDavenport:
I agree we should use this opportunity to improve the bang matching.. 
I would also like to be able to match it if the bang is at the end.. (I often after typing something I want to hit the run entry).
But I can do this later.

Calrama:
Thanks for the use case, https://github.com/DaveDavenport/rofi/pull/542/commits/da6380e364909d925882379151c71fc97d9f58fd should take care of that.

Issue #537, opened 2017-01-12
Window mode: switch to desktop needed?


lbonn:
This is a follow-up to the changes introduced by #393 and a request to consider reverting them if issues with some WMs are no longer a problem.

I use i3 with a [focus-to-last helper](https://github.com/acrisci/i3ipc-python/blob/master/examples/focus-last.py) and the switch to desktop is reported through the i3 ipc.

As a result, the script changes the focus to the previously active window on the target desktop if used after a cross-desktop switch with rofi.

## Version

1.3.1-175-g547bac0 (master)

## Launch Command

`rofi -show window`

## Steps to reproduce

* record the window focus changes of your wm
* switch to an unfocused window in another desktop with `rofi -show window`
* try to go back to the previous window by using the recorded history

DaveDavenport:
This will break awsomewm?

lbonn:
When I read the other issues about that (also #394), it seems probable...
I'll do a quick test if I have some spare time at the end of the week.

In that case, the desktop switch logic could be guarded by a `if (awesome_wm)`.

DaveDavenport:
Is there an easy way to detect awesome_wm? like property on root window?

zestyr:
FWIW, [here's how neofetch detects the current window manager](https://github.com/dylanaraps/neofetch/blob/master/neofetch#L615-L619).

PR #558, opened 2017-02-11
Window mode: intermediate switch to desktop only in awesome

lbonn:
As reported in #537, switching to the target's desktop before the final
window destination can have unintended consequences, so it's best to avoid
it if possible.

However, it is necessary inside awesome wm (#393). This commit
explicitely checks the wm's name to choose the window switching method.

### DIFF
--- a/source/dialogs/window.c
+++ b/source/dialogs/window.c
@@ -567,6 +567,47 @@ static inline int act_on_window ( xcb_window_t window )
     return retv;
 }
 
+static void switch_to_target_desktop ( xcb_window_t target )
+{
+    // Get the current desktop.
+    unsigned int              current_desktop = 0;
+    xcb_get_property_cookie_t c               = xcb_ewmh_get_current_desktop ( &xcb->ewmh, xcb->screen_nbr );
+    if ( !xcb_ewmh_get_current_desktop_reply ( &xcb->ewmh, c, &current_desktop, NULL ) ) {
+        current_desktop = 0;
+    }
+
+    // Get the desktop of the client to switch to
+    uint32_t                  wmdesktop = 0;
+    xcb_get_property_cookie_t cookie;
+    xcb_get_property_reply_t  *r;
+
+    cookie = xcb_get_property ( xcb->connection,
+                                0,
+                                target,
+                                xcb->ewmh._NET_WM_DESKTOP,
+                                XCB_ATOM_CARDINAL,
+                                0,
+                                1 );
+    r = xcb_get_property_reply ( xcb->connection, cookie, NULL );
+    if ( r && r->type == XCB_ATOM_CARDINAL ) {
+        wmdesktop = *( (uint32_t *) xcb_get_property_value ( r ) );
+    }
+    if ( r && r->type != XCB_ATOM_CARDINAL ) {
+        // Assume the client is on all desktops.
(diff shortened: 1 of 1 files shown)


source/dialogs/window.c
```diff
@@ -587,42 +628,33 @@ static ModeMode window_mode_result ( Mode *sw, int mretv, G_GNUC_UNUSED char **i
...
 
-                                                          wmdesktop,
-                                                          XCB_CURRENT_TIME );
+                wm_name = window_get_text_prop ( wm_win, xcb->ewmh._NET_WM_NAME );
+
+                if ( wm_name == NULL ) {
+                    wm_name = window_get_text_prop ( wm_win, XCB_ATOM_WM_NAME );
+                }
+
+                if ( wm_name ) {
```
sardemff7: We use GLib, you can safely go with `g_strcmp0(wm_name, "awesome") == 0`

DaveDavenport:
We currently have two different detection of WM. one for I3 (by looking if a property is there) and one for awesomeWM.

I think it might be a good idea to write a WM detection function that runs on startup. I will look at that this weekend.

lbonn:
Sure, this should be cleaner.

This PR uses a method similar to what is done by `wmctrl -m`. I tested it under i3 and awesome.

lbonn:
The wm detection that was pushed in master does not seem to work reliably under
awesome. Apparently i3 adds its name to the root but awesome does not.

Under awesome:

```
$ xwininfo -tree -root

xwininfo: Window id: 0x105 (the root window) (has no name)

  Root window id: 0x105 (the root window) (has no name)
  Parent window id: 0x0 (none)
     20 children:
     ...

     0x600007 "Awesome no input window": ("awesome" "awesome")  1x1+-1+-1
     0x600006 "Awesome systray window": ("awesome" "awesome")  1x1+-1+-1  +-1+-1
     0x600005 "awesome": ("awesome" "awesome")  1x1+-1+-1  +-1+-1
     0x600004 "Awesome WM_Sn selection owner window": ("awesome" "awesome")
     ...
```

Under i3:

```
$ xwininfo -tree -root

xwininfo: Window id: 0x105 (the root window) "i3"

  Root window id: 0x105 (the root window) "i3"
  Parent window id: 0x0 (none)
     28 children:
     ...
     0x60006a "[i3 con] workspace __i3_scratch": ("i3-frame" "i3-frame")  10x10+-15+-15  +-15+-15
     0x600068 "[i3 con] content __i3": ("i3-frame" "i3-frame")  10x10+-15+-15  +-15+-15
     0x600066 "[i3 con] pseudo-output __i3": ("i3-frame" "i3-frame")  10x10+-15+-15  +-15+-15
     0x600064 (has no name): ("i3-frame" "i3-frame")  10x10+-15+-15  +-15+-15
     0x600063 "i3": ()  1x1+-1+-1  +-1+-1
     ...
```

I believe that the call to `xcb_ewmh_get_supporting_wm_check_xxx` is necessary
to get a window with the wm's name if it exists.

DaveDavenport:
ok. will add.. 
I tried a couple wms I had installed for those it worked..

DaveDavenport:
Added, can you check.

DaveDavenport:
ok. That seems to work.

PR #564, opened 2017-02-24
Fix configuration path in main page

anthonynguyen:
`XDG_USER_CONFIG_DIR` is typically `~/.config` rather than `~/.local`


DaveDavenport:
thanks, that is a silly mistake :smile: It should have been ~/.config/

PR #574, opened 2017-03-16
Add support for multi-character separators in dmenu.c

frostidaho:
This patch adds support for dmenu-mode entry separators that are strings, instead of a single character.
I'm a neophyte c programmer, so any feedback is appreciated. Also, I'm not up to speed on how rofi does tests, so it'd also be nice if you could point me to how I should write tests for this code.

https://github.com/DaveDavenport/rofi/issues/245

### DIFF
--- a/doc/help-output.txt
+++ b/doc/help-output.txt
@@ -30,7 +30,7 @@ DMENU command line options:
 	-select [string]                       Select the first row that matches
 	-password                              Do not show what the user inputs. Show '*' instead.
 	-markup-rows                           Allow and render pango markup as input data.
-	-sep [char]                            Element separator.
+	-sep [string]                          Element separator.
 		'\n'
 	-input [filename]                      Read input from file instead from standard input.
 	-sync                                  Force dmenu to first read all input data, then show dialog.
--- a/source/dialogs/dmenu.c
+++ b/source/dialogs/dmenu.c
@@ -73,7 +73,8 @@ typedef struct
 {
     /** Settings */
     // Separator.
-    char              separator;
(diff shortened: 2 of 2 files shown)


DaveDavenport:
A few questions:
* With this patch can I still use null character ( '\0') as separator?
* What is the performance impact of this patch? Looking at it, I would not be surprised if it had an significant impact.
*  Is it still truely async? what if reading data from a pipe? can this still rofi? It looks like there are blocking calls in async functions.

frostidaho:
* It does look like using '\0' as a separator breaks something.
* I haven't profiled but there is some additional overhead with splitting some of the bigger more monolithic functions into pieces. In the case where you have a single character as a separator the code does essentially the same thing as the original version. This could probably be optimized so you wouldn't pay a penalty.
* There are a couple blocking calls in async_read_callback when a possible match is found. That needs to be fixed

DaveDavenport:
There are also extra allocations caused by the GString, I am more worried about that.

frostidaho:
Thanks - I'll keep that in mind when going over it. If my understanding of GString is correct a GString shouldn't be that much larger than a raw string in most cases. Or are you more concerned just about the number of allocations?

https://developer.gnome.org/glib/stable/glib-Strings.html#GString

DaveDavenport:
more about the extra allocations.
Anyway the only way to know is to benchmark.

DaveDavenport:
passing almost 11 month border without updates.. closing.

frostidaho:
No problem - Haven't had the time make benchmarks for it

PR #580, opened 2017-03-26
Support [hostname]:port format of .ssh/known_hosts

### LINKED ISSUE #575: ssh dialog doesn't support hosts with ssh on custom port.
## Version

Version: 1.3.1

## Configuration

debian default

## Launch Command

`rofi -show ssh`
## Steps to reproduce

* known_hosts should contain entries with port, like
`[storage.domain.my]:58319 ecdsa-...`
* select this entry in rofi

## What behaviour you see

* rofi silently closes and nothing happen

## What behaviour you expect to see

* I expect to see terminal window

sarg:
Fixes #575


DaveDavenport:
If you just 'chop off' the port number, if you want to connect again to this entry don't you connect to the wrong port? shouldn't we preserve this information.

sarg:
I agree that adding {port} placeholder to `-ssh-command` would be better and safer choice.
My solution works fine for me, because I have wildcard pattern in .ssh/config like that:
```
Host *.domain.my
	User myuser
	Port 12345
```

DaveDavenport:
Don't worry about the codecov stats.

DaveDavenport:
I often specify a different port in my ssh config file for a host.  Will this still work with this patch  as it will pass port 22 on the commandline by default?

sardemff7:
What if we added such a wonderful format string engine that you could do `${port:+-p ${port}}`? :-°
…
🚪🚶

sarg:
> Will this still work with this patch as it will pass port 22 on the commandline by default?
If the host is already in known_hosts then yes. If you type in new host without port, then no. If you type in new host with port then it will fail first but then will work. :)

How can I make `{port}` optional?

sarg:
I can set `{port}` to `-p 1234` or to empty string. Is this a good idea. Another option is to cut `{port}` from `config.ssh_command` if it is not needed. What do you think?

sarg:
I have implemented the first option. Now `{port}` will be resolved to `-p 1234` or to empty string.

sardemff7:
I’m currently porting rofi to a new format string engine, to avoid the `-p%s` trick

DaveDavenport:
Why close it? Most of this PR is still valid and needed.

mehtamanan0:
This is still a needed feature, is it still in development ?

PR #585, opened 2017-04-09
Added freedesktop icon support

AshyIsMe:
I've added application icon support to rofi for the drun mode (and drun in combi)
Uses librsvg and gtk/gdk as dependencies.

This is the most C I've written since uni so please feel free to rip it to pieces and show me where I've made a million mistakes...
See screenshots attached:
rofi-2017-04-09-1503
rofi-2017-04-09-1504

Is this a feature you would like to add to rofi?


DaveDavenport:
Thanks, the results look very nice. Will make many users very happy!

Some initial thoughts:

- GTK is a massif dependency to add.
- For me it adds 0.4 seconds to the startup time of rofi.
- It leaks memory
- If I scroll through the list it gets stuck and rofi becomes unresponsive. (need to look into this).

AshyIsMe:
Hi Dave,

Sweet, I'm glad you like it!
Totally agree about GTK.  
I think it actually only requires gdk-pixbuf 3.x but I couldn't see that in the the apt search on ubuntu 16.04.  I'll try to cut it back down to just gdk-pixbuf.
If it was just librsvg and gdk-pixbuf would that be minimal enough for you?

The startup time is likely due to the totally naive icon search in `walk_dir_for_icons()` - https://github.com/AshyIsMe/rofi/blob/master/source/dialogs/drun.c#L365 
At the moment it just does a stupid full walk of the user and system icon dirs and then /usr/share/pixmaps/.  
It just chooses the first filename match too and doesn't take size or type into consideration.
It should be possible to improve this and possibly use the DRUN_CACHE_FILE too (I haven't looked into how this caching works yet).

Memory leaks: I'm bad at C... haha.  
I'll see what I can find with valgrind memcheck.

Scrolling through the list is working for me but it could be something to do with the memory leaks, hopefully digging in with valgrind also sorts that out.

Thanks for responding, I'll fix the above issues and send an updated PR.  Probably later this week.

DaveDavenport:
I guess we need svg and png support.. So if we use cairo's png support librsvg svg support? 
I really would like to avoid gdk/gtk as dependency.  I also think the placement code should be improved.

sardemff7:
It’s gdk-pixbuf 2 (never hit version 3, the API was stable for a long time now). librsvg provides gdk-pixbuf with a launcher, so you can have SVG support without the dep, magically at runtime if librsvg is installed. But that requires an extra step so using librsvg directly can be a nice optimization.
However, gdk-pixbuf uses non-premultiplied alpha and RBGA format (not endianess-dependent) while cairo uses ARGB/BRGA (endianess-dependent). And the code to copy the data is in GDK, not gdk-pixbuf : `gdk_cairo_set_source_pixbuf()`. So the code is under LGPL v2+, and rofi is MIT. You’ll have to rewrite the full support yourself. :-)
OTOH, XDG Icon theme spec mandates 3 format, PNG, SVG and XPM. Cairo is fully able to support PNG and probably XPM without much effort. librsvg could be an optional dep (with 2 or 3 well-placed `#ifdef`).

As for getting the icon, I have an xdg-theme implementation in [libnkutils](https://github.com/sardemff7/libnkutils/blob/master/src/nkutils-xdg-theme.h), not sure how optimized it’s for rofi usage though. :-)
But you need a theme for that, I’m not sure if there is a standard way to set a theme, so we could use a new setting in rofi and just requires a theme to be set and don’t draw icons otherwise.

@DaveDavenport Did you measure the additional load time for drun only or for all modes? It could be the loading of GTK+ (which may be loading up to GL, but not sure since libepoxy can probably delay the load until real usage).

DaveDavenport:
@sardemff7  I looked at the time spend finding the icons. 
```
0.093735 (0.000006): ../source/dialogs/drun.c:get_app_icons:489 Get Desktop app icons (start)
0.102187 (0.008452): ../source/dialogs/drun.c:get_app_icons:497 Get Desktop app icons (user dir)
0.686407 (0.584220): ../source/dialogs/drun.c:get_app_icons:506 Get Desktop app icons (system dirs)
```

AshyIsMe:
The icon placement is extremely simple at the moment.  I added a tab character in to the display strings and just place the icons over that.
I figured you would want to do something much better and probably themeable with the new css changes too.

DaveDavenport:
I would strongly prefer dropping xpm support if that means dropping gtk/gdk:
 
```
diff --git a/configure.ac b/configure.ac
index 8482561..f947bc4 100644
--- a/configure.ac
+++ b/configure.ac
@@ -120,8 +120,6 @@ PKG_CHECK_MODULES([pango],    [pango pangocairo])
 PKG_CHECK_MODULES([cairo],	  [cairo cairo-xcb])
 PKG_CHECK_MODULES([libsn],    [libstartup-notification-1.0 ])
 PKG_CHECK_MODULES([librsvg],    [librsvg-2.0 ])
-PKG_CHECK_MODULES([libgdk_pixbuf], [gdk-pixbuf-2.0])
-PKG_CHECK_MODULES([libgtk], [gtk+-2.0])
 
 
 dnl ---------------------------------------------------------------------
diff --git a/source/dialogs/drun.c b/source/dialogs/drun.c
index fe559c4..b0d0bda 100644
--- a/source/dialogs/drun.c
+++ b/source/dialogs/drun.c
@@ -628,13 +628,8 @@ static cairo_surface_t *_get_icon ( const Mode *sw, unsigned int selected_line )
             /*if(surface == NULL) { g_log ( LOG_DOMAIN, G_LOG_LEVEL_DEBUG, "surface == NULL %s", dr->icon_path ); }*/
             /*return surface;*/
         } else {
-            surface = cairo_image_surface_create_from_pixbuf(dr->icon_path);
-            if(surface == NULL) {
-                g_log ( LOG_DOMAIN, G_LOG_LEVEL_DEBUG, "Icon type not yet supported: %s", dr->icon_path );
-                return NULL;
-            } else {
-                return surface;
-            }
+            g_log ( LOG_DOMAIN, G_LOG_LEVEL_DEBUG, "Icon type not yet supported: %s", dr->icon_path );
+            return NULL;
         }
     }
 }
diff --git a/source/x11-helper.c b/source/x11-helper.c
index e2b2a20..d4c1ec1 100644
--- a/source/x11-helper.c
+++ b/source/x11-helper.c
@@ -32,7 +32,6 @@
 #include <unistd.h>
 #include <stdint.h>
 #include <glib.h>
-#include <gdk/gdk.h>
 #include <cairo.h>
 #include <cairo-xcb.h>
 
@@ -160,32 +159,13 @@ cairo_surface_t* cairo_image_surface_create_from_svg (const gchar* file)
 
 cairo_surface_t* cairo_image_surface_create_from_pixbuf (const gchar* file)
 {
-    GdkPixbuf *pixbuf;
-    gint width;
-    gint height;
-    cairo_format_t format;
     cairo_surface_t *surface;
-    cairo_t *cr;
-
-    pixbuf = gdk_pixbuf_new_from_file (file, NULL);
-    if(pixbuf == NULL) {
-        return NULL;
-    }
-    format = (gdk_pixbuf_get_has_alpha (pixbuf)) ? CAIRO_FORMAT_ARGB32 : CAIRO_FORMAT_RGB24;
-    width = gdk_pixbuf_get_width (pixbuf);
-    height = gdk_pixbuf_get_height (pixbuf);
-    surface = cairo_image_surface_create (format, width, height);
-    /*g_assert (surface != NULL);*/
-    if(surface == NULL) {
+    surface = cairo_image_surface_create_from_png ( file );
+    if ( cairo_surface_status ( surface ) != CAIRO_STATUS_SUCCESS )
+    {
+        cairo_surface_destroy ( surface );
         return NULL;
     }
-    cr = cairo_create (surface);
-    gdk_cairo_set_source_pixbuf (cr, pixbuf, 0, 0);
-    cairo_paint (cr);
-    cairo_destroy(cr);
-
-    g_object_unref(pixbuf);
-
     return surface;
 }
 
```

I think it might be worth the check to see if we can do the finding/loading of icons in a separate thread. 
(not gonna be trivial).

DaveDavenport:
Would also be nice to add icon support to the window switcher.. 
Either using the passed icon name (if any) else the passed icon.

AshyIsMe:
Current Status:

- gtk/gdk dependencies removed.  Only png and svg icons supported.
- Fixed some of the memory leaks. Valgrind is still showing more leaks when compared to the source repo though.
- I expect the tests to work but Travis CI is still showing failure.   `make check` succeeds on my machine at least...

Outstanding items:
- Show icons for Window switcher mode
- Find/Load Icons on a separate thread.
- Better icon placement positioning.  Can this tie in with the new css support?

DaveDavenport:
The result for me is blurry, I guess because it seems to scale up by a factor 2.5 for most icons.

[pic](https://filebin.pw/7eQ/)

@sardemff7  has some code that implements the size threshold thingy, might be useful for this.

sardemff7:
Please see the [`wip/icons`](https://github.com/DaveDavenport/rofi/commits/wip/icons) for the libnkutils code (quite simple), done on top of your PR.
We’d need a setting for the icon theme though, for now it’s just using the `hicolor` fallback.

sardemff7:
Added (in the branch) a few improvements on loading and sizing.

sardemff7:
Cleaned everything, now in [`wip/icons-cleanup`](https://github.com/DaveDavenport/rofi/commits/wip/icons-cleanup) branch with just 3 commits full of useful stuff. :-)

sardemff7:
Merged in master, rebased manually.

PR #586, opened 2017-04-14
Handle null X events without error.  Plus Dave's Async patch

AshyIsMe:
Small patch to ignore NULL X events when xcb_connection_has_error() also reports no error.

I was seeing rofi crash without displaying at all in the wip-icons branch when running with `-dpi 1` or `-dpi 0`  like so:

```
$ G_MESSAGES_DEBUG=all ./rofi  -combi-modi window,drun -show combi -modi combi -dpi 1  > rofi.log                                                                                   
The XCB connection to X server had a fatal error: 0

$ G_MESSAGES_DEBUG=all ./rofi  -combi-modi window,drun -show combi -modi combi -dpi 0  > rofi.log                                                                                   
The XCB connection to X server had a fatal error: 0

```

The docs here suggest that if xcb_connection_has_error() returns 0 if there is no error: https://xcb.freedesktop.org/manual/group__XCB__Core__API.html#ga70a6bade94bd2824db552abcf5fbdbe3 

With this patch applied rofi runs as per normal with the `-dpi 0` or `-dpi 1` arguments.

### DIFF
--- a/include/widgets/textbox.h
+++ b/include/widgets/textbox.h
@@ -39,6 +39,7 @@ typedef struct
     guint            blink_timeout;
 
     PangoFontMetrics *metrics;
+    int              left_offset;
     //
     const char       *theme_name;
 } textbox;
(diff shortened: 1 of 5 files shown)


AshyIsMe:
I've manually merged in Dave's Async patch aswell.  It seems to work quite nicely on my machine.  Only a slight delay between when the rofi window is displayed and when the icons are displayed.  This machine has an SSD though.

Dave's patch also changed the icon placement to the far left.

PR #625, opened 2017-07-05
Fix rofi-sensible-terminal for termite.

dkasak:
When using the '-e' flag, termite expects that the program to be run and
all its arguments will be passed as a single string argument, which is
unlike some other terminals. Because of this, rofi-sensible-terminal was
broken for termite, e.g. trying to connect to a host using rofi -show ssh
doesn't open a terminal successfully.

### DIFF
--- a/script/rofi-sensible-terminal
+++ b/script/rofi-sensible-terminal
@@ -11,7 +11,18 @@
 # symlink for example.
 for terminal in $TERMINAL x-terminal-emulator termite urxvt rxvt st terminator terminology qterminal Eterm aterm uxterm xterm gnome-terminal roxterm xfce4-terminal mate-terminal lxterminal; do
     if command -v $terminal > /dev/null 2>&1; then
-        exec $terminal "$@"
+        case $terminal in
+            termite)
+                if (( $# > 0 )); then
+                    exec $terminal -e "${*:2}"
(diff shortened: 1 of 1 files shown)


DaveDavenport:
This will break when trying to start termite on my system, where I pass terminal title.
```rofi -show ssh -ssh-command '{terminal} -t "Connecting to {host}" -e {ssh-client} {host}'```


(This works btw)
```rofi -show ssh -ssh-command 'termite -t "Connecting to {host}" -e "{ssh-client} {host}"'```

Terminal arguments are really annoying, everybody does it differently.

dkasak:
Ah, right. Well, I can't think of a way to handle this cleanly then, off the top of my head.

> Terminal arguments are really annoying, everybody does it differently.

Agreed. :( 

There is a [PR for termite](https://github.com/thestinger/termite/pull/471) that wants to do away with `-e` and replace it with positional arguments. I'm hoping it could do both, by still allowing `-e` to be passed for compatibility.

DaveDavenport:
I think we had a whole discussion about this before (will look for it). In the end I just made the calls fully configurable.. not ideal, but atleast makes it work with *

DaveDavenport:
Please re-open if a good solution is found.

Issue #596, opened 2017-05-26
-no-custom option breaks -multi-select option


GrumpyDave:
Using rofi version 1.3.1 in dmenu mode.  With the -multi-select option I can choose multiple lines with shift-return.  When I add the -no-custom option shift-return exits immediately as if -multi-select was disabled.  Tested with no config file and no other options using these command lines:

```
ls | rofi -dmenu -multi-select
ls | rofi -dmenu -multi-select -no-custom
```

DaveDavenport:
Thx,

fogine:
Few days ago I downloaded [rofi-git](https://aur.archlinux.org/packages/rofi-git/) package from the aur repo.
The `-multi-select` option does not work at all.

```bash
> echo "one\nonttt\ntwo\ntwelve" | rofi -dmenu -multiselect
```
Always returns single item which matches `#window horibox listview element selected`
I realize that it's pre-release version, just want to notify about current behavior.
Thanks for sharing your work.

DaveDavenport:
works fine for me.

fogine:
I understand now, the way it works currently is that you have to explicitly "shift+enter" mark the items you want to select (I wasn't aware of it).
Currently it `-multi-select` doesn't integrate with -matching.
I'd expect items which are matched and highlighted by `show-match`  to be dumped... would be quite powerful.

PR #629, opened 2017-07-09
[Dmenu] Allow multi-select in no-custom mode in dmenu.

DaveDavenport:
Fixes: #596


gustavo-eduard:
any way to extends this feature to ssh menu too?

PR #645, opened 2017-07-31
Proof read manpages

fdmarcin:
Fix typos.
Fix American English consistency.
Fix "command line", "pop up" etc. - separate as nouns and hyphenated as compound adjectives.

closes DaveDavenport/rofi#644


doc/rofi.1.markdown
```diff
@@ -36,16 +36,16 @@ To show the run dialog:
 
 **rofi** can emulate **dmenu(1)** (a dynamic menu for X) when launched with the `-dmenu` flag.
 
-The "official" website for `dmenu` can be found [here](http://tools.suckless.org/dmenu/).
+The official website for `dmenu` can be found [here](http://tools.suckless.org/dmenu/).
```
fdmarcin: @DaveDavenport Is it official or not? If so, then dropping the quotes If not really official, then "unofficial" ;)

DaveDavenport: it is not official,  it is one of the most used versions. That is why the quotes.

doc/rofi.1.markdown
```diff
@@ -400,7 +393,7 @@ Default: *{terminal} -e {cmd}*
 
 `-run-list-command` *cmd*
 
-If set, use an external tool to generate list of executable commands. Uses 'run-command'
+If set, use an external tool to generate list of executable commands. Uses `run-command`.
 
 Default: *""*
```
fdmarcin: >Default: *""*
Should it be like this, or is something missing?

doc/rofi.1.markdown
fdmarcin: I would change to "unofficial" then, as it's a little confusing now.

doc/rofi.1.markdown
```diff
@@ -571,30 +564,30 @@ Allow multiple lines to be selected. Adds a small selection indicator to the lef
...
+Pops up a message dialog (used internally for showing errors) with *message*.
 Message can be multi-line.
 
 ### Other
 
 `-pid` *path*
 
-Make **rofi** create a pid file and check this on startup. Avoiding multiple copies running
-simultaneously. This is useful when running **rofi** from a keybinding daemon.
+Make **rofi** create a pid file and check this on startup. Avoid multiple copies running
```
fdmarcin: Should it be "avoid" as in imperative? Or something else?
If it's advice, then you could use an additional sentence explaining why, for example:
>Avoid running multiple copies, because...

doc/rofi.1.markdown
```diff
@@ -627,35 +619,35 @@ Disable parsing of configuration. This runs rofi in *stock* mode.
...
 G_MESSAGES_DEBUG=Dialogs.DRun
 ```
 
-For more information on debugging see the [wiki](https://github.com/DaveDavenport/rofi/wiki/Debugging%20Rofi)
+For more information on debugging, see the [wiki](https://github.com/DaveDavenport/rofi/wiki/Debugging%20Rofi)
 
 ## PATTERN
 
-To launch commands (e.g. when using the ssh launcher) the user can enter the used command-line,
+To launch commands (e.g. when using the ssh launcher), the user can enter the used command-line,
```
fdmarcin: Not sure what command-line means here. Do you mean "The user can enter commands in the used terminal"?

doc/rofi.1.markdown
DaveDavenport: It more stops it from accidentally happening. Maybe:  "The pid file avoids that multiple copies can run concurrently."

doc/rofi.1.markdown
```diff
@@ -665,21 +657,21 @@ The theme format below describes the old (pre version 1.4) theme format. Please
...
-  * **hlbg** Background color selected row
+  * **bg**: background color row
+  * **fg**: text color
+  * **bgalt**: background color alternating row
+  * **hlfg**: foreground color selected row
+  * **hlbg**: background color selected row
 
-The window background and border color should be specified separate. The key `color-window` contains
+The window background and border color should be specified separately. The key `color-window` contains
 a pair `background,border,separator`.
```
fdmarcin: >the key ... contains a pair `background,border,separator`.
Not a pair. Change to "set" or something else?

doc/rofi.1.markdown
fdmarcin: I think `The pid file prevents multiple copies from running simultaneously` should be fine.
Also, maybe `multiple instances`?

fdmarcin: If so, multiple copies/instances of what? rofi?

doc/rofi.1.markdown
sardemff7: The term is "tuple", but I think these args will disappear after this release on the next one, so I wouldn’t spend too much time on them. :-)

DaveDavenport: Yes. I still wonder if we should *just* mark them deprecated or completely remove them.

doc/rofi.1.markdown
DaveDavenport: instances of rofi.

doc/rofi.1.markdown
DaveDavenport: Aah default should be "{cmd}"

doc/rofi.1.markdown
DaveDavenport: hmm that is very vague yes.. 
These are keyworks that can be used in several options, like `-run-command` and '-ssh-command'.

When launching applications (for example using the *run* mode) the command executed can be specified, e.g. using the `-run-command` configuration option. The following keywords in the command will be replaced at run-time by their value.  

Not sure if this is more clear.

doc/rofi.1.markdown
fdmarcin: changing pair -> tuple
I forgot that a tuple doesn't only consist of two elements :man_shrugging:

doc/rofi.1.markdown
fdmarcin: I think I understand how `{}` keywords work, but this sentence needs paraphrasing and possibly splitting.
>To launch commands (e.g. when using the ssh launcher), the user can enter the used command-line, the following keys can be used that will be replaced at runtime:

We can just cut off the second part, and make it a separate sentence:
>The following keys can be used that will be replaced at runtime:

But I'm still unsure about what the first one is trying to say. What does it mean to "enter the used command line"? To open the shell that I use?

DaveDavenport: you can enter the 'commandline' that rofi uses the launch the application ( the '-run-command' setting). Maybe we should leave this out.

doc/rofi.1.markdown
carnager: Just get rid of the old format. rofi ships converted versions of most themes anyway.

doc/rofi.1.markdown
fdmarcin: Leaving it then.

DaveDavenport:
Thanks again for all your work. I appreciate it a lot.

fdmarcin:
Thanks! Just looking to help out :)

doc/rofi-theme.5.markdown
```diff
@@ -594,28 +593,28 @@ The following properties are currently supports:
...
 ### listview:
 * **columns**:         integer
-    Number of columns to show (atleast 1).
+    Number of columns to show (at least 1)
 * **fixed-height**:    boolean
-    Always show `lines`  rows, even if less elements are available.
+    Always show `lines` rows, even if fewer elements are available.
 * **dynamic**:         boolean
-    If the size should changed when filtering the list, or if it should keep the original height.
+    If the size should change when filtering the list, or if it should keep the original height.
```
fdmarcin: Unclear what "true" or "false" means here. I suggest "`True` if the size should change when filtering the list, `False` if it should keep the original height."

doc/rofi-theme.5.markdown
```diff
@@ -667,13 +666,13 @@ The current layout of **rofi** is structured as follow:
...
 ```
-> ci is case-indicator
+> ci is the case-indicator
 
 ### Advanced layout
 
 The layout of **rofi** can be tweaked by packing the 'fixed' widgets in a custom structure.
 
-The following widgets names are 'fixed' widgets with functionality:
+The following widgets are 'fixed' with functionality:
```
fdmarcin: The original was unclear. Did I get the meaning right?

doc/rofi-theme.5.markdown
```diff
@@ -682,22 +681,22 @@ The following widgets names are 'fixed' widgets with functionality:
  * listview
  * sidebar
 
-The following exists and automatically pack a subset of the widgets as in the above picture:
+The following exist and automatically pack a subset of the widgets as in the above picture:
```
fdmarcin: "The following exist and automatically pack a subset of the widgets as in the above picture:"
Unclear sentence.
Suggestion: "The following (options? widgets?) are available, and automatically pack a subset of the widgets as specified above:"

doc/rofi-theme.5.markdown
```diff
@@ -770,17 +769,17 @@ Explanation of the different parts:
...
-The size, on each side, of margin, border and padding can be set.
-For the border a linestyle and radius can be set.
+The size of each margin, border, and padding can be set.
+For the border, a linestyle and radius can be set.
 
 ### Spacing
 
-Widgets that can pack more then one child widget, currently box and listview, the `spacing` property exists.
-This determines the space between the packed widgets (both in horizontal as vertical direction).
+Widgets that can pack more then one child widget (currently box and listview) have the `spacing` property.
```
fdmarcin: The original was unclear. Did I get the meaning right?

doc/rofi-theme.5.markdown
DaveDavenport: Correct.  If option is set (true) the size will change while filtering the list.

doc/rofi-theme.5.markdown
DaveDavenport: Maybe this is clearer?

> The following keywords are defined and can be used to automatically pack a subset of the widgets. These originate from the default theme as depicted in the figure above.

(`originate from` or `are used in`)

doc/rofi-theme.5.markdown
DaveDavenport: is correct, but maybe maybe still add:
This property sets the spacing between the children in both the horizontal and vertical direction.

(or is picture clear enough?)

doc/rofi-theme.5.markdown
fdmarcin: done

doc/rofi-theme.5.markdown
fdmarcin: This is a very good alternative, thanks!

doc/rofi-theme.5.markdown
fdmarcin: >Widgets that can pack more then one child widget (currently box and listview) have the `spacing` property.
>This property sets the distance between the packed widgets (both horizontally and vertically).

doc/rofi-theme.5.markdown
DaveDavenport: > The following widgets are fixed and provide the following core functionality:

(what is better? core functionality? base functionality?   form (instead of provide) the core functionality of rofi)

To make clear that these are the base widgets used to implement rofi.

DaveDavenport:
I want to make a release in 2 weeks, do you think you have time to finish before that? (Otherwise I merge all the great changes so far.)

Thanks again!

fdmarcin:
Hi! I'll finish in a few days.

doc/rofi-theme.5.markdown
fdmarcin: >The following widgets are fixed, as they provide core **rofi** functionality:

fdmarcin:
Actually, I'm done @DaveDavenport :)

DaveDavenport:
Thanks Brilliant.

PR #646, opened 2017-08-01
Add new "start" matching method

Carpetsmoker:
This is similar to "normal", except that it will only match the *start* of
strings.

For example, using `rofi -show run -matching normal` and then typing `pid` on my
system shows (in this order):

	acpid
	pidgin
	pidof

But with `-matching start` it's:

	pidgin
	pidof

Many more examples where the output is much more useful, IMHO.

Could also use `matching regex` and `^pid`, but that's one extra character of
typing I need to do for ever `rofi` invocation, *and* I'd then have to worry
about escaping regexps, which I'd rather not do.

### DIFF
--- a/source/helper.c
+++ b/source/helper.c
@@ -244,6 +244,11 @@ static rofi_int_matcher * create_regex ( const char *input, int case_sensitive )
         retv = R ( r, case_sensitive );
         g_free ( r );
         break;
+    case MM_START: 
+        r    = g_regex_escape_string ( input, -1 );
+        retv = R ( g_strconcat ( "^", r, (char *) NULL ), case_sensitive );
+        g_free ( r );
+        break;
     default:
         r    = g_regex_escape_string ( input, -1 );
         retv = R ( r, case_sensitive );
@@ -588,10 +593,13 @@ int config_sanity_check ( void )
             config.matching_method = MM_FUZZY;
         }
         else if ( g_strcmp0 ( config.matching, "normal" ) == 0 ) {
-            config.matching_method = MM_NORMAL;;
+            config.matching_method = MM_NORMAL;
+        }
+        else if ( g_strcmp0 ( config.matching, "start" ) == 0 ) {
+            config.matching_method = MM_START;
         }
         else {
-            g_string_append_printf ( msg, "\t<b>config.matching</b>=%s is not a valid matching strategy.\nValid options are: glob, regex, fuzzy or normal.\n",
+            g_string_append_printf ( msg, "\t<b>config.matching</b>=%s is not a valid matching strategy.\nValid options are: glob, regex, fuzzy, start or normal.\n",
                                      config.matching );
             found_error = 1;
         }
--- a/doc/rofi.1.markdown
+++ b/doc/rofi.1.markdown
@@ -213,6 +213,7 @@ Specify the matching algorithm used.
 Current the following methods are supported.
 
 * **normal**: match the int string
+* **start**: Match at the start of the string.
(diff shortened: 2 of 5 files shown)


source/helper.c
```diff
@@ -238,6 +238,11 @@ static GRegex * create_regex ( const char *input, int case_sensitive )
         retv = R ( r, case_sensitive );
         g_free ( r );
         break;
+    case MM_START:
+        r    = g_regex_escape_string ( input, -1 );
+        retv = R ( g_strconcat ( "^", r, (char *) NULL ), case_sensitive );
```
DaveDavenport: The result of g_strconcat is leaked.

DaveDavenport:
This sounds more like a sorting then matching issue. Have you tried fuzzy match with sorting enabled. (it does fzf style sorting).
Then words starting with 'pid'  (in this case) are sorted higher then words ending on 'pid'.

I also don't think this implementation does what you want. The build in tokenization means every word in your input should start from the beginning.  Because some modes match on multiple fields, this means that every word in the input needs to match *a* field from the beginning.

Carpetsmoker:
> This sounds more like a sorting then matching issue.

Yeah, the sorting is the biggest issue.

> This sounds more like a sorting then matching issue.  Have you tried fuzzy match with sorting enabled. (it does fzf style  sorting).

I get these results with `pid`; I made sure sorting was enabled with Alt+backtick (`+` icon shows):

    pidof
    picard
    pidgin
    [..many many more...]

DaveDavenport:
(was on my mind before) maybe I can combine searching from fzf with normal matching.
that would make picard be gone from the results. (still wonder why that gets sorted above pidgin).

Carpetsmoker:
Well, for me personally as a ruke I dislike most forms "fuzzy" matching, and would prefer to just match what I type. I usually know what I want and the menu is "only" for autocomplete (like ^D in most shells).

Would you be willing to merge the PR if I fix the issues?

DaveDavenport:
How do you plan to fix the issue? (the problem with the multiple fields being matched and the tokenization).

DaveDavenport:
(p.s. I will merge a good PR)

Carpetsmoker:
> How do you plan to fix the issue? (the problem with the multiple fields being matched and the tokenization).

Not sure yet, need to investigate.

FYI I broke my right arm yesterday, and using a computer is kind of hard at the moment, so it might be a while until I have a chance to get back at this.

DaveDavenport:
> FYI I broke my right arm yesterday, and using a computer is kind of hard at the moment, so it might be a while until I have a chance to get back at this.

That is shit. Hopefully it heals well. 
(p.s. text-to-speech programming sucks. Trust me, tried it :smile:)

DaveDavenport:
How is your arm doing? hopefully everything has healed itself well.

Carpetsmoker:
It is doing well, thanks :-)

I really wanted to finish this, as rofi is kind of useless for me without it, but considering I haven't had time in the last 8 months, it's unlikely I'll have time in the next 8 months, either.

Truth is that it's been years since I programmed C and the C that I *did* do is quite different than what is used in rofi (and I never did much C programming anyway), so comparatively simple changes (like, say, figuring out how to do that `g_strconcat` properly) take me fairly long time to figure out, as I first need to (re)learn C. I've been spending most of my time in the land of Python/Go/etc.
I'd love to learn, but as always the problem is ... time :-( Being an adult can suck.

I rebased on `master`; this PR still works well for me personally. You can work on it if you feel it's worth your time, or simply close it. Either is okay with me :-)

PR #649, opened 2017-08-10
Add warning about outdated version under Ubuntu 16.04 Xenial

FlorianFranzen:
Ubuntu 16.04 packaged version of rofi is extremely outdated. This causes
all kind of weird behaviour with newer scripts, i.e. with clerk.
This commit adds a warning to the installation instructions to warn new
users about this problem and direct them to install a newer version
directly from launchpad.

Hope that this remark will save people some trouble.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -184,8 +184,13 @@ gdb rofi core
 
 ```
 apt install rofi
-
 ```
+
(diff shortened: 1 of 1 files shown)


DaveDavenport:
Maybe i3's backports for rofi is interesting to mention: http://debian.sur5r.net/rofi/.

PR #651, opened 2017-08-12
Add --pin-last option to keep the last-used entry first

CyberShadow:
This allows quickly re-running the last command by simply running `rofi` and pressing Enter.

The last commit enables the new option by default, but I can drop it at your discretion.


CyberShadow:
Updated and rebased on `next` branch.

PR #685, opened 2017-10-02
Add gentoo install instructions.

refacto:
Adding install instructions for Gentoo. The ebuild is well maintained, and one doesn't have to fuss with overlays.


DaveDavenport:
thanks.

Issue #656, opened 2017-08-19
DRUN Browser matches multiple visible and invisible fields, causing unexpected results.


BonarBeavis:
## Version

1.3.1

## Configuration

[follow instructions](https://gist.github.com/DaveDavenport/68c148d41c91186f1055f498272767a2)

## Launch Command

drun

## Steps to reproduce

* input 'fi'

## What behaviour you see

* 'Paperwork' is the first match

## What behaviour you expect to see

* 'Paperwork' should not be the first match for the search 'fi'

DaveDavenport:
DRun mode searches through many fields in the Desktop file. That might be providing a very accurate match.

BonarBeavis:
This explains a lot. Can I turn this off?

DaveDavenport:
No, this is not configurable.

BonarBeavis:
So this is intended behavior, even if it goes against the principle of least astonishment?

DaveDavenport:
That is very subjective.

If I type in 'Text editor' I expect to see all text editors, if I type in 'office' I expect to see my office tools.

The problem is there is not one solution that pleases everybody, if I try to please everybody and make 100034 options, the tool becomes unmaintainable.  (got into this mess before with GMPC that was massif and I only used 1% myself).

BonarBeavis:
I get your point, keep it simple. But then not searching all fields in the desktop entry should be implemented. I would assume, that the more general use case for drun is to launch applications and not to search for applications. 
Furthermore by including all fields the results are random for small search strings.
For example: Searching `fi` gives Steam as the first result, but searching `st` gives htop as the first result.

Maybe a solution could be: Only search the strings which are actually displayed by rofi.
Then the search for `text` still would give all text editors, but not all applications, which have `text` somewhere in their description.

DaveDavenport:
I don't agree.   rofi should **find** you what you want, so text editor -> should give all text editors.
Then the history part of rofi remembers your favorite commands and makes sure they are sorted at the top. So next time just type a few characters and your command is on top and you can quickly **run** it.


( It sounds like the functionality of the 'run' switcher matches more what you want, that just matches the executable name.)

BonarBeavis:
[...]

PR #690, opened 2017-10-04
Added option -name-only to match only desktop entry name with drun

Askrenteam:
I added the option `-name-only` to match only desktop entry name when using drun, and not description/Generic Name/etc. I've seen that other users seemed to want that behaviour ([Issue #656](https://github.com/DaveDavenport/rofi/issues/656)). By default, the usual behaviour is preserved, with the search query matching with Name/Description/Generic Name/etc. 
However, if you still feel this is unnecessary, feel free to reject this PR. If you accept it, please tell me what I can change to comply more with the style guide.
Thanks !


DaveDavenport:
Thinking about this a bit today. I will accept a patch like this if:
* It is more generic, so you can specify what fields to match. So something like:
  - all
  - name,exec
  -name,generic
  - generic,category
* better option name.  `drun-match-fields` or so.
* Add option for window mode too for consistency. (window-match-fields)
* It does not slow down.

Askrenteam:
OK, I'll be working on that !

DaveDavenport:
great!

Askrenteam:
I haven't quite gotten how to enable timings to test for possible slow-downs. Where do I need to put the flag `--enable-timings` ?

DaveDavenport:
Its a bit hacky thing, but see timing.h, you can add:

```c
TICK();
```
or 
```c
TICK_N("my timing point");
```
Around the code you want to measure, these points will be printed in the rofi-timing.log file.

source/dialogs/drun.c
```diff
@@ -547,6 +556,60 @@ static void drun_icon_fetch ( gpointer data, gpointer user_data )
     rofi_view_reload ();
 }
 
+/**
+ * Initialize desktop entry fields.
+ */
+static void drun_collect_entry_fields ( DRunModePrivateData *pd )
+{
+    unsigned int entry_fields_length = 4;
+    pd->matching_entry_fields                     = g_realloc ( pd->matching_entry_fields, entry_fields_length * sizeof ( DRunEntryField ) );
```
DaveDavenport: Is there a reason why this structure is created at runtime?  it looks like it can easily be a static global to the file.
```c
static DRunEntryField matching_entry_fields[DRUN_MATCH_NUM_FIELDS] = {
    { .entry_field_name = "name", .enabled = TRUE, },
   { },
};
```

source/dialogs/drun.c
```diff
@@ -547,6 +556,60 @@ static void drun_icon_fetch ( gpointer data, gpointer user_data )
...
+    pd->matching_entry_fields[0].entry_field_name = "name";
```
DaveDavenport: Please create an enum for the index, this helps readability. (esp. in the later part)
something like:
```c
typedef enum {
DRUN_MATCH_FIELD_NAME,
DRUN_MATCH_FIELD_EXEC,
DRUN_MATCH_FIELD_CATEGORIES,
...,
DRUN_MATCH_NUM_FIELDS
}  DRunMatchingFields;
```

source/dialogs/drun.c
```diff
@@ -547,6 +556,60 @@ static void drun_icon_fetch ( gpointer data, gpointer user_data )
...
+            gboolean matched = FALSE;
+            for ( unsigned int i = 0; i < pd->matching_entry_fields_length; i++ ) {
+                const char * entry_name = pd->matching_entry_fields[i].entry_field_name;
+                if ( strcmp ( token, entry_name ) == 0 ) {
+                    pd->matching_entry_fields[i].enabled = TRUE;
+                    matched                              = TRUE;
+                }
+            }
+            if ( !matched ) {
+                g_warning ( "Invalid entry name :%s", token );
```
DaveDavenport: Not sure yet myself? should we make this a 'stronger' error and display it to the user?

source/dialogs/drun.c
```diff
@@ -706,22 +772,30 @@ static int drun_token_match ( const Mode *data, rofi_int_matcher **tokens, unsig
             int              test        = 0;
             rofi_int_matcher *ftokens[2] = { tokens[j], NULL };
             // Match name
-            if ( rmpd->entry_list[index].name ) {
-                test = helper_token_match ( ftokens, rmpd->entry_list[index].name );
+            if ( rmpd->matching_entry_fields[0].enabled ) {
```
DaveDavenport:
```c
 if ( rmpd->matching_entry_fields[DRUN_MATCH_FIELD_NAME].enabled ) {
```

alexozer:
I also would really like a way to prioritize executable / title name matching over description. I'd argue that "finding all desktop files that match a certain category / description like 'text editor'" happens a lot less frequently than trying to open one specific program with a certain name quickly.

I run into situations like this very often:
rofi

DaveDavenport:
@alexozer Please follow the guidelines and make **a new issue**. Do not make meta issues, as they become hard to keep track off.

source/dialogs/drun.c
Askrenteam: You're right, I got confused, sorry. Not quite used to writing in C.

source/dialogs/drun.c
Askrenteam: I don't know what your policy is for this kind of problems. I got inspiration from the parsing function for the combi-modi option, so it would be coherent to keep the same type of warning in all parsing functions.

source/dialogs/window.c
```diff
@@ -369,9 +397,46 @@ static int window_match ( const Mode *sw, rofi_int_matcher **tokens, unsigned in
         }
     }
 
+    TICK_N ( "end window_match" );
     return match;
 }
 
+static void window_mode_parse_fields ()
```
DaveDavenport: this function can be executed twice is both window and windowcd are enabled. It might be good to add a guard.

Askrenteam: Isn't it the goal of the line `if ( mode_get_private_data ( sw ) == NULL )` to ensure that data aren't initialised twice ?
```c
static int window_mode_init ( Mode *sw )
{
    if ( mode_get_private_data ( sw ) == NULL ) { 
        ModeModePrivateData *pd = g_malloc0 ( sizeof ( *pd ) );
        pd->window_regex = g_regex_new ( "{[-\\w]+(:-?[0-9]+)?}", 0, 0, NULL );
        mode_set_private_data ( sw, (void *) pd );
        _window_mode_load_data ( sw, FALSE );
        window_mode_parse_fields (); 
    }   
    return TRUE;
}
static int window_mode_init_cd ( Mode *sw )
{
    if ( mode_get_private_data ( sw ) == NULL ) { 
        ModeModePrivateData *pd = g_malloc0 ( sizeof ( *pd ) );
        pd->window_regex = g_regex_new ( "{[-\\w]+(:-?[0-9]+)?}", 0, 0, NULL );
        mode_set_private_data ( sw, (void *) pd );
        _window_mode_load_data ( sw, TRUE );
        window_mode_parse_fields (); 
    }   
    return TRUE;
}
```

DaveDavenport: Yep, but different modi, with each it own private data.

Askrenteam: Oh, ok, thanks !

Askrenteam:
Do I have to add cases for the coverage tests to pass ? If so, where do I need to add them ?

DaveDavenport:
No not needed.  I might add some later .

DaveDavenport:
thanks!

Askrenteam:
Thank you, glad I could help !

PR #710, opened 2017-10-30
Add icon-size option

siiptuo:
This option is useful to prevent blurry icons. The new option is optional and
by default icons are still scaled to fill all available space.

### DIFF
--- a/config/config.c
+++ b/config/config.c
@@ -53,6 +53,8 @@ Settings config = {
 
     /** Whether to load and show icons */
     .show_icons          = FALSE,
+    /** Icon size */
+    .icon_size           = 0,
 
     /** Terminal to use. (for ssh and open in terminal) */
     .terminal_emulator = "rofi-sensible-terminal",
--- a/doc/test_xr.txt
+++ b/doc/test_xr.txt
@@ -22,6 +22,8 @@ rofi.xoffset:                        0
 rofi.fixed-num-lines:                true
 ! "Whether to load and show icons" Set from: Default
 ! rofi.show-icons:                     false
+! "Icon size" Set from: Default
+! rofi.icon-size:                      0
 ! "Terminal to use" Set from: File
 rofi.terminal:                       sakura
 ! "Ssh client to use" Set from: File
--- a/include/settings.h
+++ b/include/settings.h
@@ -70,6 +70,8 @@ typedef struct
 
     /** Whether to load and show icons  */
     gboolean       show_icons;
+    /** Icon size */
+    unsigned int   icon_size;
 
     /** Terminal to use  */
     char           * terminal_emulator;
--- a/source/view.c
+++ b/source/view.c
@@ -932,7 +932,7 @@ static void update_callback ( textbox *t, unsigned int index, void *udata, TextB
         else{
             list = pango_attr_list_new ();
         }
-        int             icon_height = textbox_get_font_height ( t );
+        int icon_height = config.icon_size ? config.icon_size : textbox_get_font_height ( t );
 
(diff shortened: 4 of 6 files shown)


DaveDavenport:
We should probably fix it so situations like this won't happen (set icon size to big):

rofi-2017-10-31-0914-00000

sardemff7:
Why do you want this option exactly? To avoid downscaling or upscaling?
We use the closest size to the target one. That means upscaling can happen if we want e.g. size 27 and both 24 and 32 icons are available, as 24 will be selected.
Changing just that should avoid most of the upscaling issues, without adding yet another option. Would that be an acceptable solution to you?
Downscaling should not be that time consuming for these extra icons, as sizes are close (it’s not like we want to downscale a 3000×3000 icon to 50×50 because we would upscale a 40×40 instead).
If it appears to be too much, we can always keep the distance calculation as-is, and prevent upscaling.

siiptuo:
@DaveDavenport Now the displayed icon size is capped.

@sardemff7 I want to prevent as much scaling as possible because many applications/themes provide pixel-perfect icons for sizes like 16x16.

sardemff7:
Still, if you have a 16×16 icon, you want to downscale it to fit the line height of e.g. 14px, or it’ll look like Dave’s screenshot.
You should actually not add a setting for size, but one for upscaling, IMO. That means for a line of 23, you’ll get the 16×16 icon not scaled, for a line of 24 you’ll get the 32×32 icon downscaled, and it should still look better than a too small 16×16 icon.
IOW: the less the user can mess with values, the better. Making semantic settings sounds saner.
Not that SVGs are still preferred, and they are rendered at the exact size they’ll be used. Such a semantic “no upscale” setting may prefer non-SVGs to use the pixel-perfect 16×16 icon. And I would rather have **1** setting than 2 just for one use case.

siiptuo:
@sardemff7 I agree that more automatic solution would be better. However simply disabling upscaling doesn't work because then icons could be displayed in different sizes and this doesn't look nice.

Alternatively icon size could be based on line height as it is now. But if the size is close to a predefined size like 16, 24 or 32, then this size is used. This way icon size is consistent but never too small.

sardemff7:
Using the “closest predefined size” has the same issue as your patch: there may be no icon of the wanted size, so these are scaled anyway.
Preferring the bigger size over the smaller one (so replacing any upscaling by a downscaling) still feel better to me.
Do you have any screenshot on how better it looks with your pixel-perfect icons over the scaled ones? (Preferably some downscaled ones, as that’s what we’d do :-P)

siiptuo:
Mostly 16x16 with no scaling
noscale

Mostly downscaled from 22x22
scale

sardemff7:
Could you try by applying the following patch to `subprojects/libnkutils` please?
```diff
diff --git a/src/xdg-theme.c b/src/xdg-theme.c
index 9b6cc51..0ef2cc3 100644
--- a/src/xdg-theme.c
+++ b/src/xdg-theme.c
@@ -871,6 +871,8 @@ nk_xdg_theme_preload_themes_icon(NkXdgThemeContext *context, const gchar * const
 static gint
 _nk_xdg_theme_icon_subdir_compute_distance(NkXdgThemeIconDir *self, gint size)
 {
+    if ( self->size < size )
+        return G_MAXINT;
     if ( self->type == ICONDIR_TYPE_FIXED )
         return ABS(self->size - size);
     if ( size < self->min )
```

DaveDavenport:
Don't want to add more options to rofi, no reply on proposed patch. Closing.

PR #720, opened 2017-11-17
Update rofi.1.markdown

ckdarby:
Fix typo


sardemff7:
@ckdarby Please, next time, don’t use such a generic "Update file" commit summary, just a "man: Fix typo" would be way better. Don’t hide the real change in the details. The details of a commit message is to explain the “Why?”. Thanks still :-)

PR #725, opened 2017-11-25
Fix errors in example themes in docs

ollien:
In 1.4.2, these themes don't compile, and I'm presented with parser errors.

This fixes that.

### DIFF
--- a/doc/rofi-theme.5
+++ b/doc/rofi-theme.5
@@ -142,7 +142,7 @@ This is a valid element name:
 .
 .nf
 
-element normal\.normal {
+#element normal\.normal {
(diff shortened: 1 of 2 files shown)


ollien:
Updated the PR slightly to highlight another fix I found for a similar problem. If you want me to squash these, that's cool, I just didn't want to force push if I didn't have to.

DaveDavenport:
In > 1.4.2 the # prefix is optional.  Why mix and match documentation of different versions and expect it to work?

ollien:
Appologies. That's on me. Thanks so much for taking the time to look at it

PR #743, opened 2017-12-23
Add some additional info on installing on 16.04

dragon788:
I went to follow the directions and they didn't let me install the newest or even prior versions of rofi until I found xcb-util-xrm higher up in the compiling instructions after reading a couple other issues. This should make it more evident to users how to get it working.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -191,7 +191,7 @@ apt install rofi
 
 **Please note that the latest version of rofi in Ubuntu 16.04 is extremely outdated (v0.15.11)** 
 
-This will cause issues with newer scripts (i.e. with clerk) and we recommend to manually download and install the deb file for zesty instead. You can find the deb on [ubuntu's launchpad page for rofi](https://launchpad.net/ubuntu/+source/rofi).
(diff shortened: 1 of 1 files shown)


balta2ar:
@dragon788 It's better to put into the docs the exact commands to install the dependencies so that users could easily copy-paste them, instead of collecting them one-by-one. Or, even better, provide a bash (sh) script to do that, e.g. `bootstrap.sh`, and mention it in the docs.

dragon788:
I agree, I was simply updating the format that was there. I can add a code block after that with the commands to install each, the downside is if they aren't `versionless` packages in the repositories, even a minor version bump will make the commands invalid.

I was looking at several PPA repositories to provide them and a couple have newer packages but they also have some other newer packages that could potentially cause issues for other applications.

Long story short I ended up getting rid of the newer rofi because of package incompatibilities and I'm waiting for Ubuntu 18.04 to get all the latest goodness. I've been looking into snaps as these are supported by a few distributions and may allow for proper bundling of potentially incompatible libraries in a manner that will allow using them without affecting the base system.

Issue #611, opened 2017-06-21
Shift left/right not recognized anymore


eigengrau:
## Version

Version: 1.3.1 - 1.3.1-645-g252d89a1 (master)

## Configuration

https://gist.github.com/anonymous/2060824de81bfb2f23d531692c503fb0

## Steps to reproduce

Using the keyboard layout loaded from [here](http://pb.gehidore.net/83Pu), press shift + alt_gr+ qwertz-s.

```
KeyPress event, serial 28, synthetic NO, window 0x4800001,
    root 0x1fa, subw 0x0, time 6258852, (824,720), root:(871,767),
    state 0x0, keycode 108 (keysym 0xfe0c, ISO_First_Group), same_screen YES,
    XLookupString gives 0 bytes: 
    XmbLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyPress event, serial 28, synthetic NO, window 0x4800001,
    root 0x1fa, subw 0x0, time 6258853, (824,720), root:(871,767),
    state 0x82, keycode 62 (keysym 0xffe2, Shift_R), same_screen YES,
    XLookupString gives 0 bytes: 
    XmbLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyPress event, serial 28, synthetic NO, window 0x4800001,
    root 0x1fa, subw 0x0, time 6260454, (824,720), root:(871,767),
    state 0x1, keycode 113 (keysym 0xff51, Left), same_screen YES,
    XLookupString gives 0 bytes: 
    XmbLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyRelease event, serial 28, synthetic NO, window 0x4800001,
    root 0x1fa, subw 0x0, time 6260593, (824,720), root:(871,767),
    state 0x1, keycode 113 (keysym 0xff51, Left), same_screen YES,
    XLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyRelease event, serial 28, synthetic NO, window 0x4800001,
    root 0x1fa, subw 0x0, time 6260794, (824,720), root:(871,767),
    state 0x83, keycode 108 (keysym 0xfe0c, ISO_First_Group), same_screen YES,
    XLookupString gives 0 bytes: 
    XFilterEvent returns: False

KeyRelease event, serial 28, synthetic NO, window 0x4800001,
    root 0x1fa, subw 0x0, time 6260830, (824,720), root:(871,767),
[...]

PR #745, opened 2017-12-25
bindings: Rework mouse bindings

sardemff7:


sardemff7:
Contains a fix for #611/#744

PR #750, opened 2018-01-11
commandline option to specify the location of the applications directory for drun mode

itspec-ru:
Added "drun-apps-dir" commandline option to specify the location of the Desktop entries for drun mode.
This option may be convenient for systems without DE (WM only) for emulation of desktop shortcuts.

### DIFF
--- a/config/config.c
+++ b/config/config.c
@@ -67,6 +67,8 @@ Settings config = {
     .run_shell_command      = "{terminal} -e {cmd}",
     /** Command executed on accep-entry-custom for window modus */
     .window_command         = "xkill -id {window}",
+    /** User's specified paths for Desktop entries */
+    .drun_apps_dir          = NULL,
     /** No default icon theme, we search Adwaita and gnome as fallback */
     .drun_icon_theme        = NULL,
     /**
--- a/doc/rofi.1
+++ b/doc/rofi.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "ROFI" "1" "December 2017" "" ""
+.TH "ROFI" "1" "January 2018" "" ""
 .
 .SH "NAME"
 \fBrofi\fR \- A window switcher, application launcher, ssh dialog and dmenu replacement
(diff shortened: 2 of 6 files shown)


DaveDavenport:
Why not just set: XDG_DATA_HOME environment variable?

sardemff7:
And why does having a DE change anything? The only DE feature in Desktop files is the `NotShowIn`/`OnlyShowIn` based on `XDG_CURRENT_DESKTOP`, which you can define too, to a value that would match nothing (e.g. `NotAThingAtAll`).

itspec-ru:
If XDG_DATA_HOME specified, then rofi wiill use entries from XDG_DATA_HOME **AND** 
XDG_DATA_DIRS. But I need get entries **ONLY** from specified directory.

If XDG_DATA_HOME=<some_dir> and XDG_DATA_DIRS=<some_dir> then I get entries **ONLY** from specified directory, but I have problems with MIME and icon themes from application. For example, pcmanfm-qt does not display the icons from the theme Breeze.

sardemff7:
You still didn’t explain the real use case behind that. That would be much more consistent to set `XDG_DATA_HOME` and have a “no apps from system” option from what you describe. And yet, maybe that use case is not worth adding an option in rofi because it may be out of its scope.
So please explain what is your (user) use case, rather than the solution(s) you came up with.

itspec-ru:
@sardemff7 

> You still didn’t explain the real use case behind that

I need to be able to control filling of list of records. It is sometimes convenient to use to display the shortcuts like on the desktop, especially if the shortcuts are non-system. Another example: I use rofi to view iptv. Now the channels are displayed in a plain list, in the future I plan to implement the list with channel logos and descriptions. For this it is convenient to use a desktop entries located in a separate directory.
> That would be much more consistent to set XDG_DATA_HOME and have a “no apps from system” 
> option

Perhaps it would be better. Difficult to come up with the name of the option than to implement it.
I already have a patch in which I used the option name "drun-desktop-mode", but it is not good.
May be "drun-no-system-dirs"?

sardemff7:
Desktop files (the ones in `applications`, that `drun` cares about) are for applications, if you have something else to list, please create a script or a plugin on your own. With a C plugin, you can perfectly have icons support and whatever. (And you can base your plugin on a list of Desktop file, but not in `applications`, just like X/Wayland sessions are Desktop files in a different directory.)

`drun` is not a generic thing, it is the implementation of a specification. Your use case *may* be legit, but I am still not sure it is in the context of `drun` itself, considering we have the DE filter already. “I need to be able to control filling of list of entries.” is still a dev use case, not a user use case. An example would be “The user need to select an application able to open a specific MIME type” (like the “Open with” menu).
You’re speaking about a “desktop mode”, maybe I can guess you are referring to the shortcuts one puts on their desktop in an old-fashioned way? In this case, I would vote to make a separate plugin, as the implementation can diverge quite a bit from a more classic applications menu (even though it would share quite some code at first). (You may gain some time by using GIO `GAppInfo`, but I’m not sure.)
We must not under estimate the cost of maintenance, even when the diff seems minimal. I know Dave is always willing to please users, but it may do more harm than good. :-)

Actually, you can set `XDG_DATA_HOME` and duplicate all Desktop files from the system to your directory with `Hidden=true` (you only need `Type=Application` and `Name=placeholder` in addition to get the file parsed).

itspec-ru:
@sardemff7 

> I am still not sure it is in the context of drun itself

You may be right. Personally, I think that the ability to define directory entries would be useful not only me. In any case I will be able to use my own patch. This pull request now not actual.
I use the logic You suggested:

> set XDG_DATA_HOME and have a “no apps from system” option

link to patch: "https://bitbucket.org/gentoo-overlay/itspec/src/eb058820e73905d7b4c9a5b521345ddcca9b2544/x11-misc/rofi/files/rofi-drun-no-system-dirs.patch?at=master&fileviewer=file-view-default"

DaveDavenport:
I am going to close this one. I don't see a general use for this. (and want to avoid add more options to the already extensive list.).

PR #761, opened 2018-02-03
Update config file location in README.md

benoliver999:
I couldn't get the configuration file to work in `~/.local` but it did work in `~/.config`. This might be distro specific but I believe it is the norm now.


DaveDavenport:
oops, indeed been wrong for 2 years :D.

Thanks.

PR #770, opened 2018-02-12
Add updated gruvbox themes

bardisty:
This adds updated gruvbox rasi themes. Mostly a refactor and small changes from the current generated versions such as padding, border size, prompt colon, and colors are easier for users to customize if they wish.

I've also opened a PR for this on the `rofi-themes` repo:
https://github.com/DaveDavenport/rofi-themes/pull/16


bardisty:
@sardemff7 Everything below the global `* { }` section is the same between themes.

DaveDavenport:
We could have an gruvbox-common.rasi that is included by every theme?
(I should check if we can include a gruvbox-common.inc, will check that later).

bardisty:
> We could have an gruvbox-common.rasi that is included by every theme?

Just pushed this. Easy revert if you guys end up not wanting to go that route.

DaveDavenport:
@carnager  Did we merge this already?

DaveDavenport:
I think this was merged another way around at some point.. if not, please let me know.

PR #795, opened 2018-04-06
[doc] theme docs not using right key for bg color

patrick-motard:
According to all of the official themes, background color is set
using the key `background-color`, not `background`. This commit
fixes references to this key in the themes markdown documentation
to use `background-color` instead.


DaveDavenport:
Thanks.  (reminder to self should also fix foreground to text-color).

PR #796, opened 2018-04-09
Add history ignore for specific programs

gergesh:
This should allow users to configure programs that won't be added to the history file.
This could be useful for programs that you wouldn't execute twice in the same manner usually, such as calculators or skipping to a specific song on your playlist.
I have to be honest here - I haven't tested this. I won't have access to a computer I can install all of the build dependencies on for a few days. The main part of it does work on its own, though.


gergesh:
Okay, sorry for taking such a long time to get to this.
At first this PR wasn't working properly, I've fixed some stuff and it now seems to work. Also sorry if it's shit C code, this is my first time ever working with it and I'm not sure what the standards are.

gergesh:
Closing in favor of an updated and cleaner PR #846

sardemff7:
You could have rebased and force-push to your PR branch :-)

PR #797, opened 2018-04-15
Create ribbon.rasi

rokit:
This is a horizontal theme like dmenu, but it's near the bottom of the screen with bigger text. It looks a lot different from any of the other options.


DaveDavenport:
Can you make a merge request for the https://github.com/DaveDavenport/rofi-themes repository.

Thanks.

Q

PR #808, opened 2018-05-16
Add alacritty to rofi-sensible-terminal

esphen:


esphen:
Sorry - I may have been a bit hasty on this PR. Alacritty requires a `-e` flag to execute a command.

Provided you would like to support alacritty, should I extend the script with an if-case? Or do you have any better ideas?

DaveDavenport:
thx.

PR #841, opened 2018-07-27
moving older emulators to end of list

jfw:
I have gnome-terminal installed, but it won't get run by this script because it finds uxterm and xterm first. Moving those to the end fixes this, and should fix the problem for other users--assuming xterm is installed by default in most distros.

### DIFF
--- a/script/rofi-sensible-terminal
+++ b/script/rofi-sensible-terminal
@@ -9,7 +9,7 @@
 # We welcome patches that add distribution-specific mechanisms to find the
 # preferred terminal emulator. On Debian, there is the x-terminal-emulator
 # symlink for example.
-for terminal in $TERMINAL x-terminal-emulator termite urxvt rxvt st terminator terminology qterminal Eterm aterm uxterm xterm gnome-terminal roxterm xfce4-terminal mate-terminal lxterminal konsole alacritty; do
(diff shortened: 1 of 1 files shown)


carnager:
there is a reason behind the order of terminal emulators. The ones that come with big DEs are explicitly listed after those that don't come automatically. The reasoning is that chances are much higher you installed "termite" or similar explicitly in order to use it.

jfw:
That's what I thought was happening--so wouldn't it make sense for xterm to
be at the end, since nearly every distro includes it by default? Or at
least after gnome-terminal, which is more specific. The order I changed it
to is the only way I could get it to load gnome-terminal instead of
xterm/uxterm.

script/rofi-sensible-terminal
```diff
@@ -9,7 +9,7 @@
 # We welcome patches that add distribution-specific mechanisms to find the
 # preferred terminal emulator. On Debian, there is the x-terminal-emulator
 # symlink for example.
-for terminal in $TERMINAL x-terminal-emulator termite urxvt rxvt st terminator terminology qterminal Eterm aterm uxterm xterm gnome-terminal roxterm xfce4-terminal mate-terminal lxterminal konsole alacritty; do
+for terminal in $TERMINAL x-terminal-emulator termite urxvt rxvt st terminator terminology qterminal Eterm aterm gnome-terminal roxterm xfce4-terminal mate-terminal lxterminal konsole alacritty uxterm xterm; do
```
sc0ttj: Also misses the excellent `mlterm` off the list.

PR #848, opened 2018-08-29
Don't expose cairo or glib in public includes

tomKPZ:
This change removes the requirement for modules
to depend on cairo and glib to build.  This change
fixes errors related to missing <glib.h> and <cairo.h>
for projects that do not wish to use these libraries.

### DIFF
--- a/include/helper.h
+++ b/include/helper.h
@@ -27,9 +27,10 @@
 
 #ifndef ROFI_HELPER_H
 #define ROFI_HELPER_H
-#include <cairo.h>
 #include "rofi-types.h"
-G_BEGIN_DECLS
+#include <stddef.h>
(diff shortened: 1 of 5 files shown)


sardemff7:
If you have missing headers, then you are not using the `rofi.pc` file, which is bad practice, and could lead in future errors that would be hard to debug.
Plugins are in no way forced to use GLib or Cairo, and duplicating their types definitions could also bring future errors (or current errors, `gsize` is `signed int` on 32bit systems).
Your change will lead to more maintenance work, to check if you didn’t add a new type in a public header that will magically break in some plugin because GLib and Cairo headers were not included.

Why do you need to get rid of these headers? They does no harm, just being around on your system, and you not using them.

PR #851, opened 2018-09-02
Fix links to forum

frioux:


sardemff7:
Good catch! Thanks for the patch :-)

DaveDavenport:
thx.

PR #853, opened 2018-09-06
Fedora build instruction

ojn:
added the list of packages needed to build on Fedora, cortesy of https://blog.wizardsoftheweb.pro/rofi-overview-and-installation/

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -197,7 +197,21 @@ These need to be manually installed before rofi can be installed either via sour
 
 ### Fedora
 
-rofi from [russianfedora repository](http://ru.fedoracommunity.org/repository)
+You can build it on Fedora 27 or 28, you would need to have these external libraries beside the build packages that were mentioned above: 
+
+```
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -197,7 +197,21 @@ These need to be manually installed before rofi can be installed either via sour
 
 ### Fedora
 
-rofi from [russianfedora repository](http://ru.fedoracommunity.org/repository)
+You can build it on Fedora 27 or 28, you would need to have these external libraries beside the build packages that were mentioned above:
```
bendem: it's not "*beside*", it's the packages that are mentioned above, the names are just different in fedora repositories. I would rephrase as

> You can follow these instructions to build rofi on Fedora, the packages name are different in the Fedora repositories:

* pango-devel
* cairo-devel
* ...

ojn:
@bendem changed it. I was unclear there, thanks for pointing that out!

sardemff7:
This is already explained right after the list, you can see that:

> On debian based systems, the developer packages are in the form of: `<package>-dev` on rpm based `<package>-devel`.

The part you are changing is more about already-built packages.

DaveDavenport:
> FYI: rofi is now in Fedora and making its way through the build system. Soon, you can just install it with `dnf install rofi`.
> 
> If you want to build it yourself, I would recommend downloading the [SPEC](https://src.fedoraproject.org/rpms/rofi/raw/master/f/rofi.spec) and then run `sudo dnf builddep rofi.spec`.

Great!, thanks.

Issue #870, opened 2018-11-02
Match commands with arguments in run mode


koppa:
## Version

Version: 1.5.1-dirty

## Launch Command

```rofi -terminal xterm -show run -matching fuzzy```

## Steps to reproduce

Enter a command with an argument, in my case: timer 5
and press Enter

## What behaviour you see

Instead of the given command ```timer 5```, only the entry testH265VideoStreamer
is shown, selected and executed.

## What behaviour you expect to see

Detect that my command is correct and match it, instead of the other command.

koppa:
Thanks, I didn't know of this feature.

Kabouik:
How would one proceed to run the custom command in a new terminal? The default binding for a new terminal is `Shift+Return`, while the default for `accept-custom` is `Control+Return`, and hitting both modifiers has no effect.

This is not a new issue, so I thought it would be best to reply to that one to ask for further details.

Kabouik:
Didn't notice #874 referenced just above. I guess this answers my question.

PR #874, opened 2018-11-06
New keybinding Ctrl-Shift-Enter

koppa:
This function combines accept-custom with the functionality to start the command in terminal.
The pull request relates to issue #870.


DaveDavenport:
I think this is a good option to add.

PR #875, opened 2018-11-13
tests: initialize char* in mode test

morxa:
Currently, test_mode_result relies on undefined behavior.
The test calls mode_result, which checks whether the pointer is NULL.
However, the pointer was never initialized, so it may or may not be
NULL, depending on the compiler.

This caused a test failure on ppc64 and Fedora 28, apparently because in
this setting, gcc sets uninitialized pointers to NULL.

By initializing the pointer to the empty string, the behavior is defined
and the test passes on all architectures.

I also started adding const qualifiers to all input `char**`, but that is more involved than I expected, so I stopped half-way. If you're interested, I can finish that up.

### DIFF
--- a/test/mode-test.c
+++ b/test/mode-test.c
@@ -135,7 +135,7 @@ END_TEST
 
 START_TEST(test_mode_result)
 {
-    const char *res = "";
+    char res[] = "";
     ck_assert_int_eq ( mode_result ( &help_keys_mode, MENU_NEXT, &res,0), NEXT_DIALOG);
     ck_assert_int_eq ( mode_result ( &help_keys_mode, MENU_PREVIOUS, &res,0), PREVIOUS_DIALOG);
     ck_assert_int_eq ( mode_result ( &help_keys_mode, MENU_QUICK_SWITCH|1, &res,0), 1);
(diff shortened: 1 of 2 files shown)


DaveDavenport:
Thanks, it should be initialized :D. 
I actually think the assert (mode.c:101) is wrong.  There should be a pointer to a pointer passed, the pointed pointer can be NULL. (ugh). The first pointer should be checked, not the second.

morxa:
Yes, I thought the same thing, but I wasn't completely sure, as both checks would make sense.

test/mode-test.c
```diff
@@ -135,7 +135,7 @@ END_TEST
 
 START_TEST(test_mode_result)
 {
-    char *res;
+    const char *res = "";
```
DaveDavenport: Not sure if const is correct here, being that the 3rd argument of mode_result is not const.

morxa: Yes, that is true, but `""` is a `const char *`, modifying it results in undefined behavior. I think the third argument of mode_result should be const, because it's an input string. (That's what I meant by I've started adding const qualifiers).

morxa: I initialized it to be modifiable. I still this is wrong, but it's consistent with the signature of `mode_result`.

morxa:
Thank you for merging!

PR #891, opened 2019-01-04
Sort drun entries alphabetically

ChrisJAllan:
Sorts drun entries alphabetically, keeps history order.

Category sorting could potentially be done by setting a per-category value on line 405, but would also need a category indicator to not be confusing.

### DIFF
--- a/source/dialogs/drun.c
+++ b/source/dialogs/drun.c
@@ -402,15 +402,7 @@ static gboolean read_desktop_file ( DRunModePrivateData *pd, const char *root, c
         pd->cmd_list_length_actual += 256;
         pd->entry_list              = g_realloc ( pd->entry_list, pd->cmd_list_length_actual * sizeof ( *( pd->entry_list ) ) );
     }
-    // Make sure order is preserved, this will break when cmd_list_length is bigger then INT_MAX.
-    // This is not likely to happen.
-    if ( G_UNLIKELY ( pd->cmd_list_length > INT_MAX ) ) {
-        // Default to smallest value.
-        pd->entry_list[pd->cmd_list_length].sort_index = INT_MIN;
-    }
-    else {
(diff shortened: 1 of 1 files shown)


source/dialogs/drun.c
```diff
@@ -579,6 +571,10 @@ static gint drun_int_sort_list ( gconstpointer a, gconstpointer b, G_GNUC_UNUSED
     DRunModeEntry *da = (DRunModeEntry *) a;
     DRunModeEntry *db = (DRunModeEntry *) b;
 
+    if (db->sort_index == da->sort_index) {
+        return strcmp(g_ascii_strup(da->name, strlen(da->name)), g_ascii_strup(db->name, strlen(db->name)));
```
DaveDavenport: this leaks memory and is not utf8 compatible.

DaveDavenport:
Fixed in latest commit. needs testing.

ChrisJAllan:
It mostly works, except the first item is seemingly random, followed by everything else being correctly sorted.

DaveDavenport:
the 'random' entries should be history. (but could have made a mistake.).

ChrisJAllan:
I have history disabled.

I looked into it, and the issue is [L412](https://github.com/DaveDavenport/rofi/blob/fbd34c963565b1cec92c10e169582b8f2dab8180/source/dialogs/drun.c#L412) trying to assign -0 to the first item. Replacing it with 
```cpp
if ( G_UNLIKELY ( nl > INT_MAX ) ) {
        // Default to smallest value.
        pd->entry_list[pd->cmd_list_length].sort_index = INT_MIN;
    }
    else {
        pd->entry_list[pd->cmd_list_length].sort_index = -nl;
}
```
fixed the issue.

DaveDavenport:
applied, thanks!!

Issue #38, opened 2014-03-29
Use only `Alt+Tab` (without pressing Enter), like in kde


pablospe:
It would be nice to use only `Alt+Tab` (without pressing Enter) to change the focus.

The proposed method would change the focus when `Alt` is not longer pressed. Once the menu is displayed, `Alt+Tab` can be used to go to the item in the menu (observe that actually only `Tab` is pressed again since `Alt` is continuously pressed). And `Alt+Shift+Tab` for the previous item. In order to cancel, `ESC` can be used (while pressing `Alt`). This is the default behavior in kde, for example.

I believe it is faster and uses less hand movements.

sanderboom:
An option to always select the MRU (most recently used) window would already be a big step :+1:

vyp:
@sanderboom Probably better left to the wm in my opinion.

roomcays:
@sanderboom isn't MRU already implemented in `window` modi? It works for me at least.

sanderboom:
@roomcays for me the current active window is selected by default. If I want 'alt-tab-like' behavior, I have to press `keydown`, `enter`.

I did a quick test with `-a 1`, but this didn't seem to work as expected. Are you using this or an other switch?

sanderboom:
@vyp The scenario I'd like to use rofi for is a 'cross workspace alt-tab'. Could probably be done using wm as well, although I think it would be a nice and non-intrusive addition to rofi.

roomcays:
I see now what you're trying to achieve. Got the same behaviour.
As a workaround you can try to emulate `rofi`, `keydown`, `enter` with [xdotool](https://github.com/jordansissel/xdotool) and [xbindkeys](http://freecode.com/projects/xbindkeys) (or alternatives).

sanderboom:
@roomcays thanks for the suggestion. In this case I'd need to use a second shortcut for this as I still want to be able to use rofi in different ways. That's why an option to preselect not the current but logical MRU item would be nice :) I'll experiment some more.

vyp:
@sanderboom Well yes, a non-cross workspace alt tab would be pretty limiting. Which window manager/de are you using? Been a long time since I've used a de, but most de's usually have some sort of alt-tab behaviour built in.

Otherwise, even if you're using a non-ewmh wm, this should be pretty easy to achieve with wmutils. But I haven't tried it myself so I'm not sure.

sardemff7:
Now that we have on-release bindings (in master), it should be possible.

[...]

PR #909, opened 2019-02-02
Port over dmenu flag '-selected-row'

maxverbeek:
In #38 people have resorted to xdotool hacks to skip to the second row on startup. Upon experimentation this seemed buggy for really fast alt tabs. This patch allows us to set the selected row using the `-selected-row` flag, just like in dmenu. Docs are incomplete so far, as I don't know where to document this. As I am also very unfamiliar with this code I have no clue if this addition is in a logical place.

For now I named this argument the same as the dmenu equivalent that I found in the docs. I don't know if this is a good idea in terms of confusion.

Feedback is welcome.

### DIFF
--- a/source/rofi.c
+++ b/source/rofi.c
@@ -195,6 +195,11 @@ static void run_switcher ( ModeMode mode )
     if ( rofi_view_get_active () == NULL ) {
         g_main_loop_quit ( main_loop  );
     }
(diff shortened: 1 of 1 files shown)


DaveDavenport:
This has been in rofi for a while. thx.

PR #911, opened 2019-02-04
Made it so the presence of a field in drun-display-format is enough t…

TheConfuZzledDude:
Made it so the presence of a field in drun-display-format is enough to parse the .desktop files for that field. Also fixed helper_string_replace_if_exists only allowing one guarded field in the display format by making the regex ungreedy


DaveDavenport:
> …o parse the .desktop files for that field. Also fixed helper_string_replace_if_exists only allowing one guarded field in the display format by making the regex ungreedy

Interesting patch. Thanks.
Is part of this message missing (where it now sais ...)

TheConfuZzledDude:
Oh yeah, that's just because I was too tired to change the pull request message from the default, it's literally just the commit message

PR #918, opened 2019-02-21
Fix RandR 1.5 check: use RANDR_PREF_MINOR_VERSION

stapelberg:
XCB_RANDR_MAJOR_VERSION/XCB_RANDR_MINOR_VERSION are the maximum supported (in code) version.

Before this commit, the check verified that your libxcb did not support a newer RandR version than your running server.
After this commit, the check verifies the server’s RandR version is at least 1.5.


DaveDavenport:
hmm that looks indeed like a silly mistake.

PR #920, opened 2019-02-22
Remove old sh-ism

frioux:
The x prefix thing hasn't been needed for quite some time, especially if you are declaring that you want to use bash.


0mp:
The removal of `x` is just a style change here. The real fix here is the change from `$@` to `$*` because you want to treat your command line arguments as one here.

PR #929, opened 2019-03-04
Refer to `sorting-method` in 1.5.2 release

dmarcoux:
Instead of `sort-method`, which is wrong.

Thank you for rofi.


DaveDavenport:
thx.

dmarcoux:
It should also be updated in the releases: https://github.com/DaveDavenport/rofi/releases

PR #934, opened 2019-03-17
List all windows, including minimized

shashurup:
I don't use a taskbar so rofi looks like a reasonable candidate to manage minimized windows. So there must be a way to make rofi show them. So here it is.


DaveDavenport:
If I try xfwm and minimize windows, they still show up in (unmodified) rofi.

What is your setup?

shashurup:
I'm currently playing with goomwwm.

DaveDavenport:
with goomwwm I get nothing back when requesting windows via 'xcb_ewmh_get_client_list`. 
I tried a few other window managers (xfwm, openbox, fvwm) they all seem to work well.

shashurup:
Currently the list of windows is taken from _NET_CLIENT_LIST property of the root window (this is what xcb does AFAIK). I'm going to double check whether it is a WM responsibility to put all windows into this property.

shashurup:
It looks like this is goomwwm problem - it is not completely EWMH compliant. https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472723904 states that there must be all windows managed by WM.

PR #950, opened 2019-04-16
Added option to disable the loading overlay in dmenu mode.

joshua-sterner:


DaveDavenport:
Hmm the 'lost' widget.. 
I think it is better if we make this part of the theme and allow it to be disabled this way.

DaveDavenport:
Will add a commit that adds it as part of the theme.
So you can disable it in the theme or via commandline:

```-theme-str 'inputbar { children: [prompt, entry, case-indicator]; }' ```

Themes can now change the placement of this widget.

Issue #941, opened 2019-03-24
rofi -show drun hides options if .desktop names repeat


unode:
## Version

% rofi -version
Version: 1.5.2

## Configuration

https://gist.github.com/unode/ddf571a9cd187fde3ca9f34030737c2f

## Launch Command

`rofi -show drun`

## Steps to reproduce

* 1 - Copy a .desktop file from a system folder (/usr/, ...) into `~/.local/share/applications` keeping its original name.
* 2 - Edit the content of the .desktop file to have a different `Name=` and `Exec=`.
* 3 - Launch rofi and see the system .desktop get ghosted by the copy
* 4 - Rename the copied .desktop to have a unique name
* 5 - Launch rofi again and the system .desktop should be visible again

## What behaviour you see

In step 3 I could only select the custom `firefox.desktop` in my home folder.
Doing step 4 fixed the issue.

## What behaviour you expect to see

All options should be displayed regardless of the name of the `.desktop` file.

DaveDavenport:
This is correct and follows the specification.
https://specifications.freedesktop.org/desktop-entry-spec/latest/

See the ID part:
https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html#desktop-file-id

```
 If multiple files have the same desktop file ID, the first one in the $XDG_DATA_DIRS precedence order is used. 
```

chabad360:
That would be fine, but if the first one is marked as hidden (i.e. by `kmenuedit`) then it wont show at all.

DaveDavenport:
I think that is exactly following the specification. Allowing you to disable items this way.

sardemff7:
I can confirm that.

chabad360:
> I think that is exactly following the specification. Allowing you to disable items this way.

True, except that in a situation where you have your shortcuts organized (cause it seems like the old behavior had it that rofi would show `Entry`, `Entry (1)`, etc.) so now if you hid the wrong one, you don't see anything...

DaveDavenport:
I am unsure what you want to us to do?

I think it is important to achieve compliance with the specification.

chabad360:
Even gnu-utils doesn't have perfect compliance, and this is **minor** non-compliance, yet it removes alot of confusion.

sardemff7:
**You** must follow the spec and name your `.desktop` file differently (and edit the `Name=` key), and it’ll magically work just the way you seem to want.

chabad360:
[...]

PR #957, opened 2019-04-30
Add an option to ignore duplicate .desktop entries [WIP]

chabad360:
Based off of the discussion from #941 this is a PR to resolve that problem.


sardemff7:
As explained on #941, this is a no-go.
We want to reduce the number of options to the minimum, and non-compliance to a well-defined and widely implemented spec is definitely not something we want.

PR #961, opened 2019-05-08
[Timings] Move into new debug system.

DaveDavenport:


sardemff7:
IIRC, `g_debug()` and friends put the newline at the end themselves, so you’ll get an empty line each time.
The rest seems good, pretty straightforward. :-)

PR #979, opened 2019-06-24
Add -active and -urgent (pythonic rows selection)

kennethho:
An easy way to mark last rows.

My use case is to place special rows (non-item, say jump points) at the end of list:
`list-items | ( cat - && printf '%s\n' jump-point-x jump-point-y)` | rofi -dmenu -active -2:-1

So they would not be in the way, yet easily reachable, e.g. hit end key

Mark last row: `-active -1`
A range: `-active 3:-2`
or a set of rows: `-active 1,0,-1`
or any combination: `-active 0,2:3,-9`

In my limited tests, they can be specified alongside the good old `-a` and `-u` counterparts, e.g.:
`rofi -dmenu -u 0 -urgent -1`

They seem general enough to be useful for people than myself, I hope.


DaveDavenport:
Interesting patch, but I don't like the addition of two extra commandline options that duplicate functionality with a slightly different syntax.
Is it possible to merge it into 1?

kennethho:
That was my initial thoughts and preference. As I looked into the the good old `-a` and `-u` before the real hacking starts, I learned that it couldn't be done in a way that maintains backward-compatibility. Considering a command in the lines of:

`rofi -dmenu -a -1 # note the negative index`

The range parsing functions would process the argument (`-1` in this case) as if it had a leading implied `0`, e.g. `-1` as if it was `0-1`

As far as I am aware of, such usage is not described (explicitly allowed) in the docs, but it works and I don't know rofi well enough to make reasonable guess on its stance on backward-compatibility. Therefore I felt I need to be conservative, don't break things that worked.

Prior to settling down on the syntax opted in the patch, a potentially more likable alternative was considered, `-A` and `-U`. But they didn't stick as soon as I realized rofi is not case-sensitive on options.

Though I can't say I thought everything through, I believe I could produce another patch that supports the original documented syntax and my proposed pythonic rows selection syntax under `-a` and `-u`. If breaking said undocumented usage is acceptable.

I am glad you found the patch interesting. Let me know what you think and how I should proceed in order to have it merged.

PR #985, opened 2019-06-29
add pythonic rows selection to -a and -u

kennethho:
Previously: #979 

> `-a` *X*
> 
> Active row, mark *X* as active. Where *X* is a comma-separated list of python(1)-style indices and ranges, e.g.  indices start at 0, -1 refers to the last row with -2 preceding it, ranges are left-open and right-close, and so on. You can specify:
> 
>   * A single row: '5'
>   * A range of (last 3) rows: '-3:'
>   * 4 rows starting from row 7: '7:11' (or in legacy notation: '7-10')
>   * A set of rows: '2,0,-9'
>   * Or any combination: '5,-3:,7:11,2,0,-9'
> 
> `-u` *X*
> 
> Urgent row, mark *X* as urgent. See `-a` option for details.

### DIFF
--- a/doc/rofi.1
+++ b/doc/rofi.1
@@ -998,13 +998,29 @@ Makes dmenu searches case\-insensitive
 \fB\-a\fR \fIX\fR
 .
 .P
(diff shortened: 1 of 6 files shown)


DaveDavenport:
@sardemff7 Any comments?

sardemff7:
I’d use the Ruby notation (`[x..y]`) but hey, patch is done and works :-D

PR #994, opened 2019-07-10
Added non-sensitivity to keyboard layout 

alexkabanov:
-matching all_kb_layouts


sardemff7:
Where is the explanation of this feature? How the user is supposed to know what it’ll do? How are we supposed to know what to review? It needs **at least** an update to the man page, and I’d appreciate a nice commit message explaining why it is necessary to add and maintain so much code. :-)

sardemff7:
You still didn’t describe “non-sensitivity to keyboard layout” at all…
Also, it sounds like something people would want *in addition* to all the already existing matching methods, which again makes it harder to implement and maintain. You **must** explain why it is important to have. (And in the commit message, feel free to squash&force-push to your PR, it’ll be nicer when merging.)

alexkabanov:
My english is bad. :)
When windows have names in two languages and when rofi launching with input language of last window, then frequently user need switch input language. 
It's slowly and uncomfortably. 
In this implementation, current device keymap used by to make regular expression. "[qй][wц][eу][rк][tе]". (keyboard layout: en,ru)
Solution, presented in issue "English as default input language", not suit because items may not only be in English. But this feature also solves the problem in issue "English as default input language". It doesn't matter what input language.

PR #1052, opened 2020-01-09
Add support for additional field on script entries `meta`

jdost:
## Purpose

For some script mode usages and displays, the content is given additional information to help in finding the desired output (like with the `rofimoji` and `kaomoji rofi` user scripts).  Since there is already a spec and support for adding icons, it seems sensible to add in a "meta" field for entries to provide matchable text but not include it in the primary entry field.

## Implementation

Following the same logic as used for parsing the `icon` field on an entry, an additional `meta` field is added and parsed out.  This field is then used along with the primary entry content when doing token matching.  This is similar in behavior to how the `drun` dialog works with matching things like `keyword` or `category` for more contextual searching.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@
 /doc/*.html
 /doc/*.doxy
 /gitconfig.h
+test/output.txt
 
 # I want to ignore log files
 *.log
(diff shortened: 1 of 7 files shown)


jdost:
Example utilization: https://gist.github.com/jdost/abbc3900e804d974777bbe62d299c4a0

DaveDavenport:
Looks good.

DaveDavenport:
Thanks.

PR #1061, opened 2020-02-21
man: update default matches for desktop files

asymmetric:
Taken from [`config.c`](https://github.com/davatorium/rofi/blob/d777d0ed4bec1fe363d35eb0b3e19a7ab17b1485/config/config.c#L120).


DaveDavenport:
thx.

Issue #1072, opened 2020-03-19
[REQUEST] allow to render multi-line text in listview entries


fogine:
It would be great if `rofi` supported rendering of multi-line strings in `listview` elements with Pango markup.  
Currently all text is stuffed into single line and a new line escape sequence `\n` is used as delimiter for new elements.  

This would enable for showing more data to user per entry.
A new option for declaring custom delimiter would have to be added.

My use case is that I need to show search results to the user where each `listview` entry has title,link and description sections which wouldn't fit into a single line (and even if it did, it would be hard to navigate through data)

`rofi 1.5.4`

DaveDavenport:
relatd to #1069

fogine:
I forgot to mention that I'd like to use it with custom `-modi` plugin that allows me to update `listview` data in realtime.

DaveDavenport:
So if #1041 is implemented this one is also resolved?

fogine:
Well, I'm reading the referenced issue again with rofi documentation.
The referenced issue talks about support for for `-sep` and `-eh` for rofi script mode which I now understand is not the same as c plugins...
It's not clear to me what is the current state of script mode.

>We need to push script modes to level of feature closer to C plugins.

I don't use the special rofi script mode currently, my intention is to use [rofi-blocks](https://github.com/OmarCastro/rofi-blocks) for more advance use cases. So it would be ideal if rofi supported `Pango` markup and line height (multi-line text) features everywhere.

DaveDavenport:
Rofi does support markup  (just about everywhere) and line heigh. So not sure where things are missing.
(no idea if rofi-blocks supports it).

fogine:
You are right, `-eh` option and pandoc markup work for c plugins.  
I got confused because at first I could not find the option for setting line height in man pages and then I saw that it's broken in the referenced issue and thought it's dmenu only option.  

I created pull request that adds the option to the man pages.

So this is no longer an issue.
This makes me excited! Thanks :)

PR #1080, opened 2020-03-31
add missing -eh (line height) option to man pages

fogine:
`rofi -h` output includes `-eh` option but `man rofi` doesnt.
adds missing `-eh` option to man pages.

related to #1072


DaveDavenport:
thanks.

PR #1098, opened 2020-04-20
mention rofi-script in rofi man page

jluttine:
Mention rofi-script in rofi man page.

NOTE: I updated the instructions on how to build the man page from markdown, I hope it was correct. However, when I built the man page, the diff was huge, almost all lines were different in some way. Perhaps some difference ronn version could cause that? Anyway, for that reason, I didn't commit the built man page but just manually picked the line that was affected by my change in the markdown file.


DaveDavenport:
Thanks, I applied it by hand an updated the instructions.

Issue #303, opened 2015-12-27
Fonts do not look sharp enough


d125q:
Is it just me or do fonts in Rofi simply not look sharp enough?  This is not very noticeable with large font sizes, but it is rather annoying with smaller fonts.  I can provide screenshots if you think it will be of any help.

DaveDavenport:
This is a side effect of using pango with cairo.
If you have a fixed size font, but request a different size it will 'scale' making it fuzzy.

d125q:
Thanks.  How can I not request a different size?

DaveDavenport:
I am not sure if you can request 'no specific' size with pango..  But just set it to the right size.

What version of rofi are you running and can you pastebin the output of `rofi -dump-xresources`. This way it won't be a complete guessing game.

d125q:
I am running the latest build from Git.  The X resources can be found on http://ix.io/n0J.  Note that the issue is present with other fonts, too.

DaveDavenport:
Hmm pragmata should be fine? can y ou show screenshot?   

`alt-shift-s` should take a screenshot and store in the XDG picture directory (if set).

d125q:
Here is a screenshot:
![rofi-2015-12-27-2249](https://cloud.githubusercontent.com/assets/7765538/12012235/2fcf7f74-acec-11e5-8575-154351fa32b8.png)

The fonts look very fuzzy compared to my regular terminal.  Perhaps fontconfig settings are not applied?

DaveDavenport:
Can you show screenshot of terminal? so I can compare. (I see a lot of aliasing in screenshot, but it looks similar as here.)
I use pango directly, I assume that handles fontconfig correctly.

d125q:
Here's an example:
![term](https://cloud.githubusercontent.com/assets/7765538/12012274/8fa08e10-aced-11e5-9758-145a8ae808ab.png)

DaveDavenport:
![compare](https://cloud.githubusercontent.com/assets/406245/12012306/77a7a40a-acee-11e5-9b70-0f054fa86bd2.png)

The green 0 is from the terminal screenshot and the other 0 from the rofi one.

DaveDavenport:
gimped:
![image2](https://cloud.githubusercontent.com/assets/406245/12012316/06343404-acef-11e5-9ea4-15b560d5966d.png)

d125q:
[...]

PR #1122, opened 2020-05-10
workaround for #303

nick87720z:
Workaround for issue #303 
Subpixel rendering may be disabled by some clip paths or when text itself is clipped. I tested only with quadrangular path. Any non orthogonal line is enough to make subpixel noop.
Workaround here is to use intermediate image surface, so that no need to disable clipping.

### DIFF
--- a/source/widgets/textbox.c
+++ b/source/widgets/textbox.c
@@ -404,6 +404,28 @@ static void textbox_free ( widget *wid )
     g_slice_free ( textbox, tb );
 }
 
+/* FIXME: workaround for cairo bug, when subpixel rendering failed with some cairo clip paths */
+static void draw_pango_layout (cairo_t * cr, PangoLayout * layout, int x, int y)
+{
+        cairo_surface_t * txt_surf;
+        cairo_t * txt_cr;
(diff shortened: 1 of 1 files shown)


DaveDavenport:
Any idea what the performance impact is of this patch?

DaveDavenport:
Got around 2% drop. Not bad.

nick87720z:
I did perf for various approaches for text rendering. Pango seems to be heaviest of all graphics, comparing to simple color fills and lines. It doesn't even honor tolerance option (not sure is it good or bad).

There is minimal example, I used for drawing.
https://gist.github.com/nick87720z/53a1c952d479b67a7f2b096a88deed81

Preserving intermediate surface/painter makes no difference for me.
Recording surface gives little speedup comparing to image.
This allowes to get rid of extent calc, but it seems that final layout draw call needs it anyway, so no diff.
cairo toy api (cairo_show_text) gave 3.5 .. 4.0x perf boost. If course it would lack pango layout support. And it gives extents only for painted area, not logical.

DaveDavenport:
thanks.

PR #1123, opened 2020-05-12
[WIP] Feature/combi format string

jchtt:
This PR adds the ability to customize the output of combi mode as described [here](https://www.reddit.com/r/qtools/comments/ghcyu4/feature_requestprototype_allow_custom_formatting/).

I tried using uncrustify but a) got an error (`unknown option 'align_number_left'`), and b) noticed that it changed a bunch of other files, so left it like this for now.

It's basically done, but I added WIP because I’m not super familiar with C (would like some code review, it’s pretty short) and I’d still have to write the documentation for it.


DaveDavenport:
I can see the use of adding `-combi-display-format`.  I am a bit weary of the other 3 options;

The option `-markup-combi` breaks entries from modi that do not use markup. If the entry contains a special character for the pango markup, it will fail to render.


I don't see the point of the linebreak options,  I think they will cause a lot of confusion.

jchtt:
Thanks for taking a look!

> The option -markup-combi breaks entries from modi that do not use markup. If the entry contains a special character for the pango markup, it will fail to render.

Ah, that’s annoying, I agree. Is there a way of escaping them if the mode does not set the style flag?

> I don't see the point of the linebreak options, I think they will cause a lot of confusion.

I admit it’s a bit tricky. The point was to be able to have a nice two line layout where both the mode and parts of the original entry can appear on the first line. That’s only possible if I can have the line break appear either in the `-combi-display-format` or in the format string of the mode in question, as in the example on reddit. It can be easily ignored if people don’t want to use it, for example in the default format string `{mode} {element}`.

DaveDavenport:
if you check if the called modi does not set the markup flag, you can use that to escape the returned string.

source/dialogs/combi.c
```diff
@@ -253,17 +250,32 @@ static char * combi_mgrv ( const Mode *sw, unsigned int selected_line, int *stat
...
+					g_free(str);
+					str = tmp_str;
+					*state |= MARKUP;
+				}
+				if ( (*state & MARKUP) && !config.markup_combi ) {
+					// Mode does use markup, but we did not want to, so escape pattern string
+					char * tmp_str = g_markup_escape_text( format_str, -1 );
+					g_free(format_str);
+					format_str = tmp_str;
+				}
```
jchtt, line 267: This should take care of the escaping.

DaveDavenport: To avoid confusion, wouldn't it be better to remove the markup_combi option and always do markup (and convert element that is not markup). 

So the combi_display_format option is always a markup string.  (don't think this is unreasonable, and how drun-display-format works)

jchtt: Done! If you think the other two options are OK, I‘ll try to put some documentation in the man page.

PR #1125, opened 2020-05-12
Bash completion: add a rofi bash-completion script

johnbeard:
Adds a basic bash completion script for the rofi(1) command.
Most parameters are handled to some extent, with some extra
logic for things like modi, windows, X11 displays, key bindings
and arguments with known options.

------

Some improvments that could be made:
* Figure out a complete keysym list (for me, scraping xmodmap is missing KP_[0-9]
* Allow to comma-join modi lists
* Maybe have a tidier way to get the option lists than scraping rofi --help or hardcoding

### DIFF
--- a/script/bash-completion/rofi
+++ b/script/bash-completion/rofi
@@ -1,0 +1,253 @@
+# rofi(1) completion                                        -*- shell-script -*-
+
+# Prints all registered modi (by parsing the help text)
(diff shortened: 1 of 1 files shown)


DaveDavenport:
We could add some specific commandline flags to rofi to get a clear list of info for auto completion. What would be needed?

PR #1126, opened 2020-05-12
Man pages: document the bang feature

johnbeard:
There's otherwise no sign of the bang feature in the man page.

If #1124 is merged, this will need a tweak: s/query begins/query contains/


DaveDavenport:
thanks.

PR #1127, opened 2020-05-15
xrmoptions: add element-height alias for eh

jaki:
When looking through other people's Rofi configuration, one might
stumble upon the option `eh`.  It's hard to tell what it does at first
glance, unlike other options.  Create an alias `element-height` for `eh`
that is more understandable.  Hopefully, people start adopting this in
their config files for readability.

### DIFF
--- a/doc/rofi.1
+++ b/doc/rofi.1
@@ -724,7 +724,8 @@ rofi \-show run \-sidebar\-mode \-lines 0
 .RE
 
 .PP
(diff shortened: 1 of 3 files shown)


DaveDavenport:
I have a strong dislike to this option and how it is implemented. Instead of adding an alias, I want to fix this right at some point.
imho it should be an option set on
```css
element-text {}
```

PR #1131, opened 2020-05-16
Support rasi config character type options

jaki:
When trying to set `matching-negate-character` in a rasi config file,
the following warning shows

> Option: matching-negate-char is not of a supported type: String.

This is because character types are not yet handled.  Handle them from
the lexer all the way to xrm options.  The format shall be single quotes
containing a single ASCII character (except newline) or certain escaped
character sequences, namely

- `\'`
- `\0`
- `\\`


DaveDavenport:
thanks.

PR #1148, opened 2020-06-17
micro-refactoring: widget.c

nick87720z:
Found more opportunity to make code more clear in small scale.


nick87720z:
Tried to make more consistency with original style. Code in widget_enable / disable looks too repeating, at least for me.

PR #1153, opened 2020-07-07
Drun - fix launchers with non-text data

nick87720z:
I still have midori webapp launcher since long time. Besides cyrillic name, it also has some non-unicode data, which could be inline resources, used by midori. Although I don't use midori anymore, there is at least example.


sardemff7:
[The spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s03.html) is clear : files are UTF-8, as in, valid UTF-8. Midori has to clean their files when generating them. If you have any other example, well, tell their upstreams to do the same.

As for the `NULL` warning, did you forget to drop a debug `printf` after debugging?

DaveDavenport:
applied to git.

PR #1154, opened 2020-07-08
fix(doc): replace misleading reference to ~/.ssh/config

jsamr:
The list of hosts in ssh mode is extracted from ~/.ssh/known_hosts according to man page, but the README inaccurately references ~/.ssh/config instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Drun mode features:
 
 ![SSH Launcher](https://davatorium.github.io/rofi/images/rofi/ssh-dialog.png)
 
-Quickly `ssh` into remote machines. Parses `~/.ssh/config` to find hosts.
+Quickly `ssh` into remote machines. Parses `~/.ssh/known_hosts` to find hosts.
 
 ## Script mode
 


DaveDavenport:
hmm that is a problem in manpage.
It by default parses the config file (known hosts hashes hosts on most systems.)

You can enable known_hosts parsting as an option.

DaveDavenport:
I just double checked, manpage is correct and readme is correct.

jsamr:
@DaveDavenport Here it is: https://github.com/davatorium/rofi/blame/next/doc/rofi.1.markdown#L502

```markdown
`-parse-known-hosts`
`-no-parse-known-hosts`

Parse the `~/.ssh/known_hosts` file for entries.

Default: *enabled*
```
I understand the “default: enabled” as ”by default, parse known_hosts file”.

DaveDavenport:
Owh. you are absolutely right, it was set to default enabled..   This is what I get adding something I don't use myself.  

Let me fix readme.

jsamr:
@DaveDavenport Oh I see! Good to see it's fixed now :-)

jsamr:
BTW, fantastic project.

Issue #1166, opened 2020-08-24
[REQUEST] Follow Type=Link standard desktop entries with drun


jsamr:
**Describe the feature**
As per version 1.5 of the [Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html), an entry can be of `Type=Link` to point to a resource, but rofi (1.5.4-2, archlinux) doesn't display such entries when running `rofi -show drun`.

Example of such entry:

```info
[Desktop Entry]
Name=Plex Server Console
Icon=plex-icon
Type=Link
URL=http://localhost:32400/web
```

**When reporting a feature request include the following information:**
- Rofi version:
  - `1.5.4`

flavorjones:
@DaveDavenport would you be open to a PR introducing this behavior?

DaveDavenport:
Yes.  I guess best is to launch URL  with `xdg-open` util.

flavorjones:
I've opened #1168 with a proposed implementation for this.

PR #1168, opened 2020-08-25
Follow Type=Link standard desktop entries with drun

flavorjones:
__Summary__

This PR implements a solution for feature request #1166, which asked for support for "Link" type desktop entries.

__Context__

The freedesktop spec allows for "Link" type desktop entries, which contain a "URL" attribute. This is sometimes used for packages like the debian package for Folding@Home which runs a local web server for administrative functions.

__Implementation Details__

I've implemented what I think is the least-invasive solution to this problem, which is to:

- add an enum, `DRunDesktopEntryType`, representing the desktop entry type (following the names and values in the freedesktop spec)
- add a member to `DRunModeEntry` representing the desktop type
- sanity check "Application" and "Link" types separately (one requires "Exec" and the other requires "URL")
- new method `launch_link_entry` to invoke `xdg-open` with the URL
- have `drun_mode_result` check the desktop entry type and conditionally call `launch_link_entry`

Some decisions which could be revisited include:
- branching on desktop entry type inside `exec_cmd_entry` instead of the caller `drun_mode_result`
- assembling a command string in `launch_link_entry` that's then taken apart by `helper_execute_command` -- we could easily just pass an array of args around with a little additional complexity

Some things that I think make sense to implement later, only if folks ask:
- allowing the `xdg-open` command to be overridden via environment variable or configuration
- allowing URL to be a matching field in DRun searches

I'd love to hear your feedback, and am happy to edit the PR in response.

### DIFF
--- a/config/config.c
+++ b/config/config.c
@@ -125,6 +125,8 @@ Settings config = {
     .drun_show_actions         = FALSE,
     /** Desktop format display */
     .drun_display_format       = "{name} [<span weight='light' size='small'><i>({generic})</i></span>]",
+    /** Desktop Link launch command */
+    .drun_url_launcher         = "xdg-open",
 
     /** Window fields to match in window mode*/
     .window_match_fields       = "all",
--- a/config/config.c
+++ b/config/config.c
@@ -116,14 +116,16 @@ Settings config = {
     .tokenize        = TRUE,
     .matching        = "normal",
     .matching_method = MM_NORMAL,
-    /** Desktop entry fields to match*/
+
+    /** Desktop entries to match in drun */
     .drun_match_fields = "name,generic,exec,categories,keywords",
+    /** Only show entries in this category */
     .drun_categories   = NULL,
-    /**  Desktop format display */
-    .drun_display_format       = "{name} [<span weight='light' size='small'><i>({generic})</i></span>]",
     /** Desktop entry show actions */
     .drun_show_actions         = FALSE,
-    /** Desktop entry show actions */
+    /** Desktop format display */
+    .drun_display_format       = "{name} [<span weight='light' size='small'><i>({generic})</i></span>]",
+
     /** Window fields to match in window mode*/
     .window_match_fields       = "all",
     /** Monitor */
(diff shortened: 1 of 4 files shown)


source/dialogs/drun.c
```diff
@@ -57,10 +57,20 @@
 
 #define DRUN_CACHE_FILE            "rofi3.druncache"
 #define DRUN_DESKTOP_CACHE_FILE    "rofi-drun-desktop.cache"
+#define XDG_OPEN_COMMAND           "xdg-open"
```
DaveDavenport, line 60: Should we make this a config option?

flavorjones: I think it makes sense to, absolutely. I didn't include it in this PR because I was looking for a small initial contribution. I'm happy to give it a whirl today. WDYT about `-drun-launcher` as the config option name?

flavorjones: Added to the PR a `drun-launcher` config option.

config/config.c
```diff
@@ -116,14 +116,18 @@ Settings config = {
...
     .drun_categories   = NULL,
-    /**  Desktop format display */
-    .drun_display_format       = "{name} [<span weight='light' size='small'><i>({generic})</i></span>]",
     /** Desktop entry show actions */
     .drun_show_actions         = FALSE,
-    /** Desktop entry show actions */
+    /** Desktop format display */
+    .drun_display_format       = "{name} [<span weight='light' size='small'><i>({generic})</i></span>]",
+    /** Desktop Link launch command */
+    .drun_launcher             = "xdg-open",
```
DaveDavenport, line 129: should it be drun_url_launcher, as it is not used for launching applications?

DaveDavenport:
It looks good (I wonder if we should call it `drun-url-launcher` to make clear it is only used for urls and not applications.).

I will probably still add this to the upcoming 1.6 release.

flavorjones:
OK, I've updated the PR to use `drun-url-launcher`.

flavorjones:
Hang on - I can see at least one more place where code is assuming `.exec` is non-NULL, let me fix it and look a bit deeper.

flavorjones:
OK, I've force-pushed some fixes, namely:

- ensure that `exec` is never uninitialized
- make sure `drun_token_match` doesn't match against `exec` if it's NULL

PR #1173, opened 2020-08-30
Rename SORT_FZF to SORT_FUZZY

MaskRay:
The fuzzy matching algorithm used to be selected by `-matching fuzzy`.  A
refactoring changed it to `-sorting-method fzf` but incorrectly attributed the
algorithm to unrelated "fzf". Rename to SORT_FUZZY and keep `-sorting-method
fzf` for a while for compatibility.

### DIFF
--- a/include/settings.h
+++ b/include/settings.h
@@ -46,7 +46,7 @@ typedef enum
 typedef enum
 {
     SORT_NORMAL = 0,
-    SORT_FZF    = 1
+    SORT_FUZZY  = 1
 } SortingMethod;
 
 /**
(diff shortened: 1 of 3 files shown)


DaveDavenport:
```
The fuzzy matching algorithm used to be selected by -matching fuzzy. A
refactoring changed it to -sorting-method fzf but incorrectly attributed the
algorithm to unrelated "fzf". Rename to SORT_FUZZY and keep -sorting-method fzf for a while for compatibility.
```

I don't follow..   sorting and matching are not the same in rofi (as mentioned in the issue) . 

Matching algorithm is set here: https://github.com/davatorium/rofi/blob/next/source/helper.c#L189

Sorting is set: https://github.com/davatorium/rofi/blob/next/source/view.c#L647

Issue #483, opened 2016-10-16
Can't input words in chinese or japanse with fcitx


theFool32:
## Version

1.2.0

arch+bspwm
## Configuration

[configuration](https://gist.github.com/theFool32/d4deb49c7a0da1d0023010a04d5d6deb)
## Launch Command

rofi -show run -lines 3
## Steps to reproduce

launch rofi and type words.
## What behaviour you see

can only input english letter. however, pasting chinese works.
## What behaviour you expect to see

works with chinese.
Thx. :smiley:

DaveDavenport:
@sardemff7  ideas?

Can you give more information about how to reproduce this? I never type chinese or heard of fcitx.

theFool32:
@DaveDavenport 
[fcitx](https://fcitx-im.org/) is an input method framework which i used to type chinese and japanse.
You can reproduce this by [installing fcitx](https://wiki.archlinux.org/index.php/Fcitx) and try to type chinese in rofi.

Or are there any successful example in typing other language?
Thank you.

ps. my locale setting here.
```
> locale
LANG=zh_CN.UTF-8
LC_CTYPE="zh_CN.UTF-8"
LC_NUMERIC="zh_CN.UTF-8"
LC_TIME="zh_CN.UTF-8"
LC_COLLATE="zh_CN.UTF-8"
LC_MONETARY="zh_CN.UTF-8"
LC_MESSAGES="zh_CN.UTF-8"
LC_PAPER="zh_CN.UTF-8"
LC_NAME="zh_CN.UTF-8"
LC_ADDRESS="zh_CN.UTF-8"
LC_TELEPHONE="zh_CN.UTF-8"
LC_MEASUREMENT="zh_CN.UTF-8"
LC_IDENTIFICATION="zh_CN.UTF-8"
LC_ALL=
```

DaveDavenport:
Yes, I had people test different languages with different scripts before.

DaveDavenport:
I can confirm it does not work. Unsure why though, I have 0 knowledge about this :(

sardemff7:
Did it use to work?

On the Arch wiki, I see this export: `export XMODIFIERS=@im=fcitx`, does it help? I think not but it’s not hard to test.

AFAICT, rofi never supported XIM, nor any input method. That would need a whole new extension to support, so a lot of code.
[...]

PR #1174, opened 2020-08-30
Add a fast fuzzy filter to avoid g_regex_match in SORT_FUZZY mode

MaskRay:
Based on #1173

### DIFF
--- a/include/helper.h
+++ b/include/helper.h
@@ -238,6 +238,15 @@ gchar *rofi_escape_markup ( gchar *text );
  * @param pattern   The user input to match against.
  * @param plen      Pattern length.
  * @param str       The input to match against pattern.
+ *
+ * @returns whether the pattern is a case-insensitive subsequence of str.
+ */
+gboolean rofi_scorer_fuzzy_filter ( const char *pattern, glong plen, const char *str);
+
+/**
+ * @param pattern   The user input to match against.
+ * @param plen      Pattern length.
+ * @param str       The input to match against pattern.
  * @param slen      Length of str.
  *
  * FZF like fuzzy sorting algorithm.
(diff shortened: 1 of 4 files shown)


DaveDavenport:
Is this utf-8 compatible?

MaskRay:
Is there a way typing case-significant Unicode? :) (#483)

DaveDavenport:
How is what you can enter relevant for sorting?  the entered value is not sorted. :smile:

MaskRay:
Made it utf-8 clean

DaveDavenport:
I see for fuzzy match you do not use `mode_token_match ( t->state->sw, t->state->tokens, i );` anymore?   This will break matching for different modi.

MaskRay:
I don't know how to run rofi under cgdb, so it is a bit difficult for me to understand the logic. `mode_token_match` appears to call g_regex_match, which may be slow when running in dmenu mode on a collection of thousands of lines.

If `mode_token_match` is still significant, we can run `mode_token_match` after the filter.

DaveDavenport:
`If mode_token_match is still significant, we can run mode_token_match after the filter.`

That would not produce the same result in all cases.

MaskRay:
> `If mode_token_match is still significant, we can run mode_token_match after the filter.`
> 
> That would not produce the same result in all cases.

Why? Can you elaborate on what `mode_token_match` does?

DaveDavenport:
`mode_token_match` is forwarded to the active mode, that uses this to match against (possibly) invisible fields.

In f.e. drun you can select what fields to match/display independently.
In f.e. dmenu you can set an invisible item to match against (on request).
in rofi-blezz this is used to override the matching .

MaskRay:
> And as mentioned in other MR, sorting and matching are separate items in rofi. This will break that.

I think I simply don't have enough context to fix the issues. For me, I just need a somewhat usable fuzzy matching algorithm.... don't want to deal with separate matching and sorting which does not fit well to me. I'll close this PR then.

PR #1199, opened 2020-09-17
Remove extra colons

The-Compiler:
Rofi already adds a colon after the prompt given with -p, so the "monkey:" example would actually lead to a "monkey::" prompt.


DaveDavenport:
thx

PR #1208, opened 2020-10-05
added command line argument to only show local user desktop files

APotOfSoup:
I found that pruning my desktop files that belong to the entire system to shorten my rofi drun list was proving difficult.

I've added another command line argument 
    -drun-local-only

When this argument is passed, it only reads the desktop files from the user and skips the walk of system directories.

That way the desktop files purposely placed in the ".local/share/application" folder are the only ones shown in the menu.

I hope I got the man page updates correct. Let me know if there is some documentation I'd need to update that I missed

### DIFF
--- a/doc/rofi.1
+++ b/doc/rofi.1
@@ -373,6 +373,14 @@ Default: *true*
 .RE
 
 .PP
+\fB\fC\-normalize\-match\fR
+
+.PP
+Normalize the string before matching, so o will match ö, and é matches e.
+This is not a perfect implementation, but works.
+For now it disabled highlighting of the matched part.
+
+.PP
 \fB\fC\-no\-lazy\-grab\fR
 
 .PP
@@ -518,6 +526,12 @@ Default: false
 
 .fi
 .RE
+
+.PP
+\fB\fC\-drun\-local\-only\fR
+
+.PP
+Only load desktop files from user local data directory.
 
 .PP
 \fB\fC\-window\-match\-fields\fR \fIfield1\fP,\fIfield2\fP,...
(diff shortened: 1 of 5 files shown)


DaveDavenport:
This was already possible by setting XDG_DATA_DIRS to an empty place.

```
XDG_DATA_DIRS="/tmp" rofi -show drun
```

APotOfSoup:
I did originally set the XDG_DATA_DIRS to an empty directory but that causes other issues with the menu, and it also passes that environment variable on to the program launched.

For example, with the XDG_DATA_DIRS, I have a GTK icon theme set and a Rofi theme that shows icons. It will show all icons that are in my ".local/share/icons" pack, but icons are missing if they don't have a match in that local icon set.

xdg

If the XDG_DATA_DIRS not set, the icons that exist in /usr/share/icons will still be found and applied.

non-xdg

Here is an example with these changes running `rofi -show drun -theme clean-drun-local-only`

working

Notice how the programs are all in the same locations as the first picture but all the icons show. Comparing to the second picture you can also see the avahi program is not there as it is not in my local desktop files.

With passing XDG_DATA_DIRS to the program selected in the prompt, I ran into an issue that the selected program, in my case `gpick` failed to open when i run rofi using `XDG_DATA_DIRS="/tmp" rofi -show drun` but will run if rofi is run using `rofi -show drun`

Here is the terminal output after `gpick` fails to load.
failed


Because of these issues I looked to make this change so that I woudn't have to set XDG_DATA_DIRS to skip the system desktop files.

Let me know if you need more info.

DaveDavenport:
Did not think of that, thx.

DaveDavenport:
You can now do using the new config option:
```css
configuration {
   drun {
      /** Parse user desktop files. */
      parse-user:   true;
      /** Parse system desktop files. */
      parse-system: false;
   }
}
```

To tweak this behavior.

(a commandline also works:   `rofi -theme-str 'configuration { drun { parse-system: false; }}' )

PR #1214, opened 2020-10-14
[WIP] [FileBrowser] Support symbolic links

progfolio:
The new file browsing does not show symlinks. This pull adds basic support. I'm not a c programmer and I did the bare minimal change to get it working on my end, so I've marked this a WIP.

Thanks,
Nicholas Vollmer

### DIFF
--- a/source/dialogs/filebrowser.c
+++ b/source/dialogs/filebrowser.c
@@ -132,13 +132,15 @@ static void get_file_browser (  Mode *sw )
                 case DT_UNKNOWN:
                 case DT_SOCK:
                     break;
+                case DT_LNK:
                 case DT_REG:
                 case DT_DIR:
                     pd->array = g_realloc ( pd->array, (pd->array_length+1)*sizeof(FBFile));
                     // Rofi expects utf-8, so lets convert the filename.
                     pd->array[pd->array_length].name = g_filename_to_utf8 ( rd->d_name, -1, NULL, NULL, NULL);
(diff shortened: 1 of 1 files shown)


source/dialogs/filebrowser.c
```diff
@@ -132,13 +132,15 @@ static void get_file_browser (  Mode *sw )
...
+                case DT_LNK:
                 case DT_REG:
                 case DT_DIR:
                     pd->array = g_realloc ( pd->array, (pd->array_length+1)*sizeof(FBFile));
                     // Rofi expects utf-8, so lets convert the filename.
                     pd->array[pd->array_length].name = g_filename_to_utf8 ( rd->d_name, -1, NULL, NULL, NULL);
                     pd->array[pd->array_length].path = g_build_filename ( cdir, rd->d_name, NULL );
-                    pd->array[pd->array_length].type = (rd->d_type == DT_DIR)? DIRECTORY: RFILE;
+                    pd->array[pd->array_length].type = ( (rd->d_type == DT_DIR) || (rd->d_type == DT_LNK) ) ?
+                                                        DIRECTORY: RFILE;
```
progfolio, line 143: This is obviously incomplete as we're always treating a symbolic link as a directory. So it will fail when one selects a symlink which points to a regular file.

progfolio: I take it I need to `stat` symlinks to see if theyr'e regular files or directories?
Tried that approach in latest commit.

DaveDavenport:
Thanks.  Fixed this.

Issue #600, opened 2017-05-27
[Feature request] Option to highlight option under mouse


gandalf3:
When [using rofi with the mouse](https://github.com/DaveDavenport/rofi/issues/441#issuecomment-304430763), it'd be nice to have more visual feedback on which option the mouse is over, to make it easier to click fast without fear of missing. Maybe an option to select/highlight the option under the cursor?

Kommynct:
I can't even make it clickable, am I missing something?

oredaze:
I would also like to see this feature, it is very handy when using single click to accept. If not, alternatively make it so that before the single click activates an entry, the entry is briefly highlighted before it executes, because now it is a bit ambiguous what you have clicked.

xlucn:
Is this already implemented by the merged PR above?

DaveDavenport:
It sais 'merged' so yes.

DaveDavenport:
See: https://github.com/davatorium/rofi/releases/tag/1.7.0

Shamaoke:
So, how to configure rofi to highlight an entry under the mouse pointer?

DaveDavenport:
See manpage. Please use discussion forum for support questions.

PR #1234, opened 2020-12-09
Added -hover-select option that automatically selects the entry under the cursor

rahulaggarwal965:
This pull request addresses #600 and adds the `-hover-select` config/command-line option to select the entry currently under the mouse.

 I did not make this option change any existing key-bindings. I tried to include relevant documentation, and the implementation should be reasonably efficient (no forcing unneeded redraws).


elzii:
Bump - would love to see this merged!

PR #1237, opened 2020-12-17
add prefix matching method feature

unisgn:
With the normal matching method, most of the time, there are too many results, which means more key press is needed to get what I want. With the fuzzy matching method, even more results!
So I made this prefix matching method, which can **_quickly narrow down_** the query results, help me to get the target result **_as less key press as possible_**.
Of course, add a '\b' before each keyword with regex matching method can also do the work, but apparently I will not likely to type extra same keys before each keyword.
Hope it useful for others.

### DIFF
--- a/source/helper.c
+++ b/source/helper.c
@@ -179,6 +179,18 @@ static gchar *fuzzy_to_regex ( const char * input )
     return retv;
 }
 
+static gchar *prefix_regex (const char * input)
+{
+    gchar *r = g_regex_escape_string(input, -1);
+    GString *str = g_string_new(r);
+    g_string_prepend(str, "\\b");
+    g_free(r);
+    char *retv = str->str;
+    g_string_free(str, FALSE);
+    return retv;
+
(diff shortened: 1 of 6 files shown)


source/helper.c
```diff
@@ -181,14 +181,7 @@ static gchar *fuzzy_to_regex ( const char * input )
...
 {
-    gchar *r = g_regex_escape_string(input, -1);
-    GString *str = g_string_new(r);
-    g_string_prepend(str, "\\b");
-    g_free(r);
-    char *retv = str->str;
-    g_string_free(str, FALSE);
-    return retv;
-
+    return g_strconcat("\\b", g_regex_escape_string(input, -1), NULL);
```
DaveDavenport, line 184: this is a memory leak.

unisgn: Oops, not familiar with C, made a fix.

DaveDavenport:
thanks.

PR #1289, opened 2021-03-24
Update README.md

zero77:
This quickly shows all package versions across different distros at a glance. 
Which lets you decide whether your distros is upto date or you need to build for sauce to update, which is useful when trying to fix known bugs.


DaveDavenport:
thx

DaveDavenport:
hmm that svg is bigger then I expected, is there a smaller version?

zero77:
> hmm that svg is bigger then I expected, is there a smaller version?

I am not aware of a smaller one but, i will open a pr if i find one.

PR #1302, opened 2021-04-15
Specify requisite OS

Gitoffthelawn:
Clearly specify the target OS in the README.

Thank you for Rofi.  If Rofi is compatible with other operating systems as well, they can be specified too. :)


DaveDavenport:
not sure this is correct. It will run fine on BSD too.

Gitoffthelawn:
@DaveDavenport Thanks. Per your feedback, I just added BSD.

DaveDavenport:
I find it a bit weird to add this to the title.  Rofi does not really focus on a certain platform, it is more the OS providing the right elements to run rofi. Parts of rofi might work on windows with the right elements 

I have added to the README that rofi is known to work on linux and bsd.

PR #1308, opened 2021-05-07
Add 'small caps' to 'Text Style' format in manpage

Cj-bc:
'small caps' is listed in 'Text Style.IP' section in 'rofi-theme.5' and also
[lexer treat it](https://github.com/davatorium/rofi/blob/413eddd825e7c5bc6699352c989d994a40b357c1/lexer/theme-lexer.l#L212), but this wasn't in 'format' regex shown there.

Therefore, I thought it's missing, added it.

P.S. I'm a bit confused by output of 'go-md2man' as it seems to modify more than what I changed, but I committed it as is.

### DIFF
--- a/doc/rofi-theme.5
+++ b/doc/rofi-theme.5
@@ -1,3 +1,4 @@
+.nh
 .TH ROFI\-THEME 5 rofi\-theme
 .SH NAME
 .PP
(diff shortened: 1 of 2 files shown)


DaveDavenport:
I kept it out of the manpage because I noticed it did not work reliable on most of my systems. (no impact). 
I added it in hope pango would fix it later.
It works for you?

Cj-bc:
When I read through the man page, I felt a bit confused because 'small caps' is listed below, but not in the format.
especially, I didn't know how it should be written in config so I made this PR

not working neither...
I think small caps is depends on fonts, but I don't know which font does support it.

DaveDavenport:
removed it.

PR #1313, opened 2021-05-16
Add cursor property

TonCherAmi:
This pull request adds a new theme property `cursor` that allows specifying the type of mouse cursor that is set when the widget is hovered. It can be set to either `default`, `pointer`, or `text`:

- `default` is the default arrow cursor
- `pointer` could be used for clickable areas:
image
- `text` could be used for input fields:
image


DaveDavenport:
Nice. 
For me the textbox does not work and the side-bar buttons are not changed.
I wonder if we should make it  a property on the widget itself.

TonCherAmi:
> For me the textbox does not work and the side-bar buttons are not changed.

Huh, interesting.. Could you please share your config?

> I wonder if we should make it a property on the widget itself.

Yeah, I guess that would be ideal and I would personally prefer it that way, although it would take more effort to implement it that way and I'm not sure how useful it would end up being really.

TonCherAmi:
Actually, it doesn't look that difficult to do, I'm going to give it a shot.

DaveDavenport:
I ran it with -no-config. 
```rofi  -no-config -show run -sidebar-mode```

TonCherAmi:
> I ran it with -no-config.
> `rofi -no-config -show run -sidebar-mode`

Yep, the sidebar buttons do not work. However the input box seems fine on my system. I think the issue might be that the cursor name for the I-beam on yours is not in the array. I'll try searching for some more alternative names, they seem to vary from theme to theme so it's a bit difficult to get them right.

DaveDavenport:
I never set cursor themes anymore..

(long long long ago I did write a small ui to select them, gcursor, but that got absorbed into gnome settings at some point (and removed again)).

TonCherAmi:
> I never set cursor themes anymore..

I think it's not cursor themes specifically, but rather GTK themes in general.

I've mostly got widget properties working, though there's a small problem with the recently added `-hover-select` option since it relies on the fact that `listview_element_motion_notify` is called conditionally and we now must call it unconditionally so that the cursor is updated for every widget. How do you feel about removing that option and making selection on hover the default behavior?

DaveDavenport:
no direct feeling..  don't use mouse with rofi myself. Will give it some thoughts.

TonCherAmi:
> no direct feeling.. don't use mouse with rofi myself. Will give it some thoughts.

Nevermind.. We can just check the settings inside the handler, no need to remove it.

DaveDavenport:
Looks good.

DaveDavenport:
thanks.`

PR #1316, opened 2021-05-22
Minimize rebuilds on git version change

nick87720z:
For now full rebuild happens when current commit changes, which may happen without actual content change - e.g. something is commited or uncommited via `git reset` without `--hard` option. This change makes only necessary files being rebuild - those, using GIT_REVISION.


DaveDavenport:
@sardemff7  you added that option, any feedback on this.

I never had issues with this (But rofi compiles so quick, I would not have noticed anyway).

sardemff7:
Seems like a good move, but you should put the `#include` next to the nkutils one, in an `#else` branch for line 50 `#ifdef`

Issue #1264, opened 2021-01-25
[BUG] Dashed border feature doesn't appear to work


adrian5:
## Version

1.6.1

## Configuration
```markdown
Detected modi:
    * window
    * windowed
    * run
    * ssh
    * drun
    * +combi
    * keys
    * file-browser

Compile time options:
    * window enabled
    * drun enabled
    * gcov disabled
    * asan disabled
```
#### config.rasi

```css
configuration {
  modi: "combi";
  combi-modi: "drun,window";
  terminal: "termite";
  theme: "test-theme";
  lines: 5;
  fixed-num-lines: false;
  display-combi: "? ";
  display-drun: "r";
  display-window: "w";
}
```

#### test-theme.rasi

```css
#inputbar {
  border: 2px dash;
}
```

## Launch Command

```shell
rofi -show combi
# i3: exec --no-startup-id rofi -show -combi
```

## Steps to reproduce

Invoke rofi.

## What behaviour you see

Any border is displayed as a solid line, instead of "dashed". It looks identical with `border: 2px solid;` and `border: 2px;`.

## What behaviour you expect to see

I was trying to replicate the look seen in e.g. the [Adapta-Nokto](https://github.com/davatorium/rofi-themes/tree/master/Official%20Themes#adapta-nokto) theme.

DaveDavenport:
thx.

DaveDavenport:
It seems the new border code dropped support for this.

Not sure how easy to is to fix in current design. @nick87720z

adrian5:
Not sure how important it is either. Solid (i.e. regular border) is probably sufficient, just need to reflect that change in the docs. I reckon it's better to mention the removal, than to remove any mention; for situations like mine.

DaveDavenport:
I did not realize it broke, lets see if we can fix it.

nick87720z:
@DaveDavenport Probably you will do it faster, because I still did not try to lean yacc/lex languages - analysing for patterns in existing codes seems to be useless there, as I did when implemented 'mm' support.

DaveDavenport:
the yacc/lex has this already,  it mostly the drawing code that needs to be fixed..

DaveDavenport:
Fixed thanks to @nick87720z .

PR #1317, opened 2021-05-23
Fix line drawing

nick87720z:
Fixed anti-linestyle regression, introduced by commit e52094ee7a2c216ddf74660a0b4a2e51d7433948 ( #1264 )
There were much more commits, but probably this time as usually not good idea to mix different things in one originally dedicated branch.


DaveDavenport:
I see a small regression:
on next:
rofi-2021-05-23-1512-00002
on this patch.
rofi-2021-05-23-1512-00001

DaveDavenport:
Just after merging (I did do some tests) I hit a corner case I think:

image

On left it looks fine, on right it has some black parts.

DaveDavenport:
image
bit better visible.

PR #1323, opened 2021-06-05
Fix tests.

TonCherAmi:
These two fail to link for me.

https://travis-ci.org/github/davatorium/rofi/builds/773144200


DaveDavenport:
thanks, did not have time to fixe them yet.

PR #1325, opened 2021-06-05
[FileBrowser] Allow setting startup directory

### LINKED ISSUE #1259: [REQUEST] Ability to provide a startup directory for the integrated file-browser mode
## Before creating a feature request

- [x] I checked the *next* branch to see if the feature has already been
  implemented

- [x] I searched existing reports to see if it is already requested.

## What is the user problem or growth opportunity you want to see solved?

I'd like to be able to specify a path to a startup directory that would be used by the file-browser mode that was integrated into rofi in v1.6.1. This would allow for (for example) always starting the browser in a consistent location such as the user's home directory, rather than starting the browser in the last location. This would be set either in the user's config file or as a runtime option, e.g. `rofi -modi file-browser -show file-browser -browser-start-dir /home/user`.

## How do you know that this problem exists today? Why is this important?

The issue was mentioned a month ago by someone on the reddit forum [here](https://www.reddit.com/r/qtools/comments/kchx8c/open_directories_with_filebrowser/), and one of the authors (Qball Cow) responded saying to make a feature request -- however it seems that the original poster has not yet done so.

The man pages (as far as I can tell) have not yet been updated to cover the file-browser mode, the output of `rofi --help | grep -i browser` shows the mode is only mentioned as one of the detected modi, and searching the github repo didn't turn up any documentation that I could find either. So if the feature already exists, then it is currently well-hidden.

## Who will benefit from it?

Anyone using the file-browser mode in rofi who wants a consistent startup location, which could be set contextually. For example, using one keypress to open it in the user's home directory, and a different keypress to open in a work directory etc.

## Version

Version: 1.6.1

## Configuration

Output of `rofi -help`: [gist here](https://gist.github.com/dave-woods/821e3f9c0ca0e6b4693c76c83a92e1ac)

**Additional details:**
- Include a link to your private config: not really relevant for mine, but [here](https://gist.github.com/dave-woods/8c276bad9f5af12b9a978e1ff7497e41)
- Include screenshots/casts of your issue: n/a

TonCherAmi:
Adds a new option - `-file-browser-directory` that allows specifying the file browser startup directory.

Closes #1259

### DIFF
--- a/config/config.c
+++ b/config/config.c
@@ -128,6 +128,9 @@ Settings config = {
     /** Desktop Link launch command */
     .drun_url_launcher         = "xdg-open",
 
+    /** Directory the file browser starts in */
+    .file_browser_directory = NULL,
+
     /** Window fields to match in window mode*/
     .window_match_fields       = "all",
     /** Monitor */
--- a/doc/rofi.1
+++ b/doc/rofi.1
@@ -1181,6 +1181,19 @@ Set ellipsize mode to start. So end of string is visible.
 Pops up a message dialog (used internally for showing errors) with \fImessage\fP\&.
 Message can be multi\-line.
 
+.SS File browser settings
+.PP
+\fB\fC\-file\-browser\-directory\fR \fIdirectory\fP
+
+.PP
+Directory the file browser starts in.
+
+.PP
+.RS
+
+.fi
+.RE
+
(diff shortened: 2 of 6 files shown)


DaveDavenport:
Thanks!

PR #1352, opened 2021-07-10
[FileBrowser] Add option to group directories ahead of files.

TonCherAmi:
Adds a new config option `directories-first` that makes it possible to toggle grouping of directories before files. It is set to `true` by default.


DaveDavenport:
thanks.

PR #1401, opened 2021-08-30
Fix typo

tuurep:
This text can appear in window mode listing, it has a confusing typo. Commit fixes typo.


DaveDavenport:
thanks.

PR #1472, opened 2021-09-22
[Textbox|Filebrowser] Display filenames with invalid UTF-8 characters

### LINKED ISSUE #1471: [BUG] Crash in file-browser when displaying filenames with special characters.
### Rofi version (rofi -v)

1.7.0

### Configuration

https://gist.github.com/Adamskye/99150e9f7a82a5f7f7dee6a52d5dc014#file-roficonfigs-txt

### Launch command

rofi -show filebrowser

### Step to reproduce

1. Download this [brokenfolder.zip](https://github.com/davatorium/rofi/files/7210726/brokenfolder.zip). Inside should be an empty folder with special characters in its filename.
2. Extract the zip file and place the folder somewhere.
3. Run `rofi -show filebrowser`and navigate to the directory that contains the broken folder.
4. Rofi won't be able to see the filename and will crash instead.

If you try and run `rofi -show filebrowser`again, rofi will crash immediately since it's still inside the directory containing the broken folder. Editing/deleting the cache or deleting the folder will allow you to run the rofi file-browser again.

### Expected behavior

This is what I expect the filebrowser to look like when viewing the folder's name:
expected

### Actual behavior

It crashes upon attempting to render the folder's name. In the terminal, I get:
`zsh: segmentation fault (core dumped) rofi -show filebrowser`.

### Additional information

Rofi does not crash when using file-browser-extended, a different file-browser plugin, rather than rofi's built-in file-browser. I used file-browser-extended to get the screenshot in the expected behaviour section.

Adamskye:
Convert filenames to UTF-8 in such a way that it won't crash while still attempting to display the filename to some extent. Therefore, it will be possible to differentiate between filenames even if they contain invalid characters.

Created to fix #1471.

### DIFF
--- a/source/dialogs/filebrowser.c
+++ b/source/dialogs/filebrowser.c
@@ -248,11 +248,7 @@ static void get_file_browser(Mode *sw) {
             g_realloc(pd->array, (pd->array_length + 1) * sizeof(FBFile));
         // Rofi expects utf-8, so lets convert the filename.
         pd->array[pd->array_length].name =
-            g_filename_to_utf8(rd->d_name, -1, NULL, NULL, NULL);
-        if (pd->array[pd->array_length].name == NULL) {
-          pd->array[pd->array_length].name =
-              g_strdup("Invalid UTF-8 filename.");
-        }
+            rofi_force_utf8(rd->d_name, strlen(rd->d_name));
         pd->array[pd->array_length].path =
             g_build_filename(cdir, rd->d_name, NULL);
         pd->array[pd->array_length].type =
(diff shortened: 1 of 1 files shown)


DaveDavenport:
What is the performance impact of this patch?

Also why remove the actual locale conversion?  This is still needed. 
https://developer.gimp.org/api/2.0/glib/glib-Character-Set-Conversion.html#g-filename-to-utf8

Adamskye:
I was going to amend my commit so it only used `rofi_force_utf8` only if it contained invalid characters, but it seems that you've beaten me to it and done the same thing. I'm going to close this pull request then as well as the issue.

PR #1478, opened 2021-09-26
[rofi-sensible-terminal] Use /bin/sh instead of /usr/bin/env bash

jirutka:
This script is fully POSIX-sh compatible, i.e. it's portable and can be run in any shell, not just bash. It's incorrect to assume that bash is available on every *nix system (it's not), but /bin/sh is.

### DIFF
--- a/script/rofi-sensible-terminal
+++ b/script/rofi-sensible-terminal
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 #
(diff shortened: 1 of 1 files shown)


metov:
The change to `sh` seems like a good idea, but isn't it better to make it `/usr/bin/env sh`?

jirutka:
> isn't it better to make it /usr/bin/env sh?

No, `/bin/sh` is de-facto standard supported by all *nix systems.

PR #1545, opened 2021-12-14
modi return value 

numberokta:
hi, when using modi, rofi return nothing. example:
`ret=$(rofi -modi "test:$HOME/test" -show test)`
**$ret is null**
it would be great if you add an option that return value, something like this:
`echo -en '\0return\x1fvalue\n'`
or when script return nothing, rofi return last script arguments **$@**


DaveDavenport:
What is this? a request in a huge PR? 
Closing this now as this is not something I can do anything with.

numberokta:
sorry, my bad, wrong place to request

PR #1570, opened 2022-01-22
Add -combi-display-format

jirutka:
This PR implements a new option `-combi-display-format` (analogous to `-drun-display-format`) that allows to change position appearance of the mode name in the combi entries.


source/dialogs/combi.c
```diff
@@ -35,6 +35,7 @@
 #include <stdlib.h>
 
 #include "mode-private.h"
+#include "widgets/textbox.h"
```
DaveDavenport, line 38: Why is this include included?

jirutka: It defines the `MARKUP` constant.

DaveDavenport: aah, we should fix that, so we don't need to include widget.

Its been a while since I look at this, how this work if a row does not have MARKUP set but contains a character that needs to be escaped.. f.e. `&`

DaveDavenport:
```
(process:2032982): Pango-WARNING **: 18:56:41.978: pango_layout_set_markup_with_accel: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity — escape ampersand as &amp;
```

DaveDavenport: yep it breaks.

jirutka: To be honest, I don’t know how this will interact with entries from the modes that don’t use markup. If I don’t set `MARKUP` here, entries from `drun` are formatted, but entries from `ssh` are not. Maybe we should escape `str` if it originates from a mode that doesn’t use `MARKUP` (so the text may contain characters that need to be escaped), but how to detect this?

DaveDavenport: the same way you set it.

jirutka: Hmm, I think that I know the way…

DaveDavenport: check if it is set in `state`.

jirutka: Okay, I just update the commit.

jirutka: I misunderstood the code a bit, there’s actually no need for extra `entry_state`.

DaveDavenport:
thanks,

(we should probably deprecate the hide-combi-mode-prefix (or how it is called)).

jirutka:
> we should probably deprecate the hide-combi-mode-prefix (or how it is called)

That’s a good idea!

PR #1571, opened 2022-01-24
Add support for setting tabs stops on textboxes

jirutka:
This allows to emulate a multi-column layout inside the text boxes.

I’m not happy with using the Padding struct for this, but I didn’t find any sensible way how to handle property with multiple distance values. Moreover, the latest Pango release (1.50) added support for tab alignments other than left (right, center and decimal) which would allow even more interesting layouts.

**Example:**

```sh
rofi -modi combi -show combi -combi-modi drun,ssh -combi-display-format '{text}&#09;{mode}'
```

```css
element-text {
  tab-stops: 500px;
}
```

rofi

### DIFF
--- a/doc/rofi-theme.5.markdown
+++ b/doc/rofi-theme.5.markdown
@@ -216,6 +216,7 @@ The current theme format supports different types:
  * an orientation
  * a cursor
  * a list of keywords
+ * an array of values
  * an environment variable
  * Inherit
 
@@ -532,6 +533,13 @@ Specify the type of mouse cursor that is set when the mouse pointer is over the 
 
 A list starts with a '[' and ends with a ']'. The entries in the list are comma-separated.
 The `keyword` in the list refers to an widget name.
+
+## Array of values
+
+* Format: `{ value, value, ... }`
+
+An array starts with a '{' and ends with a '}'. The entries in the array are comma-separated.
+Currently, only the integer and distance types are supported as array values.
 
 ## Environment variable
 
--- a/doc/rofi-theme.5.markdown
+++ b/doc/rofi-theme.5.markdown
@@ -451,6 +451,22 @@ style property.
 
 > When no unit is specified, pixels are assumed.
 
+
+## Tab Stops
+
+* Format: `{Integer}`
+* Format: `{Distance}`
+* Format: `{Distance} {Distance}`
+* Format: `{Distance} {Distance} {Distance}`
+* Format: `{Distance} {Distance} {Distance} {Distance}`
+
+You can specify location of up to four tab stops by their distance from the beginning of the line.
+Each distance must be greater than the previous one.
+The text appears to the right of the tab stop position (other alignments are not supported yet).
+
+If no unit is specified, pixels are assumed.
+
+
 ## Position
 
 Indicate a place on the window/monitor.
@@ -768,6 +784,7 @@ The following properties are currently supported:
 * **placeholder-color**: Color of the placeholder text.
 * **blink**:             Enable/Disable blinking on an input textbox (Boolean).
 * **markup**:            Force markup on, beware that only valid pango markup strings are shown.
+* **tab-stops**:         Set the tab stops.
 
 ### listview:
 * **columns**:         integer
--- a/doc/rofi-theme.5.markdown
+++ b/doc/rofi-theme.5.markdown
@@ -452,21 +452,6 @@ style property.
 > When no unit is specified, pixels are assumed.
 
 
-## Tab Stops
-
-* Format: `{Integer}`
-* Format: `{Distance}`
-* Format: `{Distance} {Distance}`
-* Format: `{Distance} {Distance} {Distance}`
-* Format: `{Distance} {Distance} {Distance} {Distance}`
-
-You can specify location of up to four tab stops by their distance from the beginning of the line.
-Each distance must be greater than the previous one.
-The text appears to the right of the tab stop position (other alignments are not supported yet).
-
-If no unit is specified, pixels are assumed.
-
-
 ## Position
 
 Indicate a place on the window/monitor.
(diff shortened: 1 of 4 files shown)


DaveDavenport:
Very nice!  thanks.
Would it be worth it to  extend the theme parser?

jirutka:
> Would it be worth it to extend the theme parser?

I think so. I’ve wanted to change the parser, but I haven’t found a way to do it without fundamentally modifying it or introducing some weird self-purpose syntax for writing tab stops. Even now the syntax for the border properties is hacked into the “padding” type, so I wanted to avoid another abuse of this type.

If I hadn’t considered these limitations, I would have suggested the following grammar (in [EBNF](http://www.cl.cam.ac.uk/~mgk25/iso-14977.pdf) notation):

```
align = "left" | "right" | "center" | "decimal"
tab-stop = distance, [ align ]
tab-stops = tab-stop, { ",", tab-stop }
```

Examples:
```
tab-stops: 50em;
tab-stops: 200px center, 300px, 500px right;
```

jirutka:
What I would _ideally_ like is true multi-column layout inside the element-text, i.e. if the text before the tab is wider than the tab size, it does not “push” the text after it to the next tab stop, but shorten it with ellipsis. But this would be too complicated to implement.

DaveDavenport:
adjusting the parser is not that hard, what we mostly miss is a generic 'list' type.

DaveDavenport:
I can try to extend the parser so the list type is generic.

DaveDavenport:
Added a 'testing'  set type to git.

```css
element-text {
   tabs:  { 100px, 200px, 300px, 100 em };
}
```

should be accepted now. It is fully untested.

DaveDavenport:
Some useful accesors needs to be written for it still.

jirutka:
> Added a 'testing' set type to git.

How do you define the “set”? Set is typically defined as an _unordered_ collection of _unique_ values. If we stick “each distance must be greater than the previous one”, then it can work, but it seems very strange. I’d expect a list, not a set.

Wouldn’t it be better to generalize the existing list syntax (`[ 100px, 200px, 300px, 100em ]`)? Also, how would you implement support for the tab alignment?

DaveDavenport:
The '[]' syntax was what I wanted to do, but that would break the existing one and would severely complicate the code to get it to work.   'Set' name is not the best name, but the list name was taken..  In the code you get a list. (see the commit msg).

DaveDavenport:
If this works well, I will see if I can find a work-around for it, but did not want to spend that effort for nothing.

(If it makes you happy I can do a sed rename to array?)

jirutka:
> (If it makes you happy I can do a sed rename to array?)

Yeah, array would be more correct and less confusing name. I’m gonna update this PR.

DaveDavenport:
Done.. its something intended to be temporary anyway.

jirutka:
Something is wrong, I’m getting `NULL` from `rofi_theme_get_array_distance` and I dunno why yet.

DaveDavenport:
I just added a test, that worked.

DaveDavenport:
```c
rofi_theme_parse_string("element-text  { tabs: { 10, 20px, 30px, 40px };}");
  ck_assert_ptr_nonnull(rofi_theme);
  // ck_assert_ptr_null ( rofi_theme->widgets );
  ck_assert_ptr_null(rofi_theme->properties);                                                                                                                                            
  ck_assert_ptr_null(rofi_theme->parent);                                                                                                                                                
  ck_assert_str_eq(rofi_theme->name, "Root");                                                                                                                                            
  GList *l = rofi_theme_get_array_distance(&wid, "tabs");                                                                                                                                
                                                                                                                                                                                         
  ck_assert_int_eq(g_list_length(l), 4);                                                                                                                                                 
                                                                                                                                                                                         
  int i = 10;                                                                                                                                                                            
  for (GList *iter = g_list_first(l); iter != NULL; iter = g_list_next(iter)) {                                                                                                          
    RofiDistance *d = (RofiDistance *)iter->data;                                                                                                                                        
    ck_assert_int_eq(d->base.distance, i);                                                                                                                                               
    i += 10;                                                                                                                                                                             
  }                                                                                                                                                                                      

  g_list_free_full(l, g_free);                     
  ```
that worked.

**Comment by jirutka** [comment]
It works when I do the following change:

```diff
 GList *rofi_theme_get_array_distance(const widget *widget,
                                      const char *property) {
- ThemeWidget *wid2 = rofi_theme_find_widget(widget->name, widget->state, TRUE);
+ ThemeWidget *wid2 = rofi_theme_find_widget(widget->name, widget->state, FALSE);
```

Every other accessor in `theme.c` passes `FALSE` to this function.

DaveDavenport:
What is your theme entry?

jirutka:
What do you mean by theme entry?

DaveDavenport:
what did you set in your theme you are trying to match?

jirutka:
Aha, this:

```css
element-text {
    tab-stops: { 100px, 200px };
}
```

DaveDavenport:
Yeah, looking at it it should probably be FALSE.  
This is still (partly) inherited to how we did inheriting in the past (people found inheriting confusing).  

I should really clean this up.

DaveDavenport:
updated.

doc/rofi-theme.5.markdown
```diff
@@ -450,6 +450,22 @@ style property.
...
+
+## Tab Stops
+
+* Format: `{Integer}`
+* Format: `{Distance}`
+* Format: `{Distance} {Distance}`
+* Format: `{Distance} {Distance} {Distance}`
+* Format: `{Distance} {Distance} {Distance} {Distance}`
+
+You can specify location of up to four tab stops by their distance from the beginning of the line.
```
DaveDavenport: Do we still have the 4 limit?

jirutka: No, I just updated the docs.

jirutka:
> Done.. its something intended to be temporary anyway.

What will be the longer-term solution?

DaveDavenport:
My plan is (but I need to give it some more thought) to merge 'list' and 'array'.
So it is:
```css
tabs: [10px,20px,30px];
```
but I need to see how I resolve the element names here, but I think I have a solution that keeps things clean-ish.

DaveDavenport:
The `{}` was for you to be able to continue, while i thought about this one.

jirutka:
What about tab alignments (left, right, center, decimal)? I see this as the main motivation for changing the syntax, otherwise “padding” would be ok-ish.

DaveDavenport:
We can add that now more easily, but I dont want to update to pango 1.5 yet (I have no machine with that yet)
.

jirutka:
> We can add that now more easily

:+1:

>  (I have no machine with that yet)

Me neither, it’s not even in Alpine Linux Edge yet, but we are already working on it. It will be in upcoming v3.16 which should be released between May and June.

DaveDavenport:
Thanks!

PR #1572, opened 2022-01-24
Merge List and Array type into one.

DaveDavenport:


DaveDavenport:
@jirutka  I think this is sufficient to merge '{}' and '[]' into one.

This is more flexible then needed, we can  have any property between []'s.

lexer/theme-parser.y
```diff
@@ -581,14 +577,10 @@ t_property_element
         $$ = rofi_theme_property_create ( P_COLOR );
         $$->value.color = $1;
 }
-| T_LIST_OPEN t_property_element_list_optional T_LIST_CLOSE {
+| T_LIST_OPEN t_property_element_set_optional T_LIST_CLOSE {
```
jirutka: This is probably wrong…

source/view.c
```diff
@@ -2011,12 +2022,17 @@ RofiViewState *rofi_view_create(Mode *sw, const char *input,
...
   GList *list =
-      rofi_theme_get_list(WIDGET(state->main_window), "children", "mainbox");
-  for (const GList *iter = list; iter != NULL; iter = g_list_next(iter)) {
-    rofi_view_add_widget(state, WIDGET(state->main_window),
-                         (const char *)iter->data);
+      rofi_theme_get_list_strings(WIDGET(state->main_window), "children");
+  if (list == NULL) {
+    rofi_view_add_widget(state, WIDGET(state->main_window), "mainbox");
+  } else {
+    // rofi_theme_get_list(WIDGET(state->main_window), "children", "mainbox");
```
jirutka: What’s the purpose of this “comment”?

lexer/theme-parser.y
```diff
@@ -306,8 +304,6 @@ static ThemeColor hwb_to_rgb ( double h, double w, double b )
 %type <ival>           t_property_highlight_styles
 %type <ival>           t_property_highlight_style
 %type <ival>           t_property_line_style
-%type <list>           t_property_element_list
-%type <list>           t_property_element_list_optional
 %type <list>           t_property_element_set
 %type <list>           t_property_element_set_optional
```
jirutka: Hmm, shouldn’t be `t_property_element_set` removed instead of `t_property_element_list`?

source/view.c
DaveDavenport: forgot to remove.

lexer/theme-parser.y
DaveDavenport: just a naming thing

lexer/theme-parser.y
DaveDavenport: renaming thing,  forgot the 2nd one. fixed now.

lexer/theme-parser.y
jirutka: Naming is important…

jirutka:
I just tested it and it works fine. :+1:

DaveDavenport:
thanks.
I could make some more cleanups based on this, but need to see how to do that without breaking syntax ;-/

PR #1599, opened 2022-02-23
Minor corrections to 'modi' verb and '#' mention

RayZ0rr:
>
> - [x] This PR targets the `next` branch and not `master`
> - [x] Its commits' summaries are reasonably descriptive
> - [x] You've described what this PR addresses below
> - [x] You've included links to relevant issues, if any with `#issue_num`
>

Add some mentions of `modi` in readme and mention of `#` as a modi separator

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ install **Rofi**.
 
 #### Running rofi
 
-To launch **rofi** directly in a certain mode, specify a mode with `rofi -show <mode>`.
+To launch **rofi** directly in a certain [modi(modes)](#modi), specify a mode with `rofi -show <modi>`.
 To show the `run` dialog:
 
     rofi -show run
@@ -136,7 +136,9 @@ If no modes are specified, all configured modes will be enabled.
 To only show the `run` and `ssh` launcher:
 
     rofi -modi "run,ssh" -show run
+The list of modes can also be `#` separated. The following is same as the above command:
 
+    rofi -modi "run#ssh" -show run
 
 The modi to combine in combi mode.
 For syntax to `-combi-modi`, see `-modi`.
@@ -152,7 +154,7 @@ mkdir -p ~/.config/rofi
 rofi -dump-config > ~/.config/rofi/config.rasi
 ```
 
-This creates a file called `config.rasi` in the `~/.config/rofi/` folder. You can modify this file to set configuration settings and modify themes. `config.rasi` is the file rofi looks to by default.
+This creates a file called `config.rasi` in the `~/.config/rofi/` folder, which is one of the default config file rofi checks for. You can modify this file to set configuration settings and modify themes.
 
 Please see the [configuration guide](https://github.com/davatorium/rofi/blob/next/CONFIG.md) for a summary of configuration options. More detailed options are provided in the manpages.
 
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ install **Rofi**.
 
 #### Running rofi
 
-To launch **rofi** directly in a certain [modi(modes)](#modi), specify a mode with `rofi -show <modi>`.
+To launch **rofi** directly in a certain [modi(mode)](#modi), specify a mode with `rofi -show <modi>`.
 To show the `run` dialog:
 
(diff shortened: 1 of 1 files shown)


DaveDavenport:
This seems incorrect..  You launch one mode, and you enable multiple modi.

DaveDavenport:
The '#' is an exception for help with i3, it is mentioned in the manpage. I don't see why this should be added to the README.

RayZ0rr:
> This seems incorrect.. You launch one mode, and you enable multiple modi.

The correction is to indicate that `mode` and `modi` will be used interchangeably so that the user won't be confused. Like here from README:
> To launch rofi directly in a certain mode, specify a mode with rofi -show <mode>. To show the run dialog

So, that users won't be confused by `mode`/`modi` terminology.

RayZ0rr:
> The '#' is an exception for help with i3, it is mentioned in the manpage. I don't see why this should be added to the README.

Other than i3, this can also be useful for some sort of user-script utilities. In case someone misses it in the readme.

DaveDavenport:
your correction is still wrong, `rofi -show <mode>` takes one mode, not multiple.
`rofi -show <modi>`  (that you even indicates is  "modes", so the fact it is multiple was not confusing) is not valid!

For lists comma is a logical separator, the fact we support `#` is just to workaround i3 weirdness, I see no reason why somebody should use this in a script.   I also find 'some sort of user script' no clear explanation why it has added value as it has no example or more information.

RayZ0rr:
> your correction is still wrong, `rofi -show <mode>` takes one mode, not multiple. `rofi -show <modi>` (that you even indicates is "modes", so the fact it is multiple was not confusing) is not valid!

There was a small typo. I have corrected it. However, it should be clear from what I said before what I was trying to achieve.

> The correction is to indicate that mode and modi will be used interchangeably so that the user won't be confused. Like here from README

Also, I know that it takes one argument only.
> `rofi -show <mode>` takes one mode, not multiple

If the typo was causing confusion then I have fixed it. There's no other location I have mentioned it takes multiple arguments.

> For lists comma is a logical separator, the fact we support `#` is just to workaround i3 weirdness, I see no reason why somebody should use this in a script. I also find 'some sort of user script' no clear explanation why it has added value as it has no example or more information.

This may or may not be added. It's rather trivial I guess

DaveDavenport:
I don't understand the miscommunication,  this edit makes it worse imho. 

I am not sure how to explain,

The show command takes one 'mode' name, not multiple. saying `rofi -show modi` is therefor wrong!

mode -> single
modi -> plural

DaveDavenport:
> This may or may not be added. It's rather trivial I guess

I want a good reason why it should be added, I prefer not to give superfluous information in the readme. It is mentioned in the manpage, the only reason to use the illogical # as list separator is for i3 users and it explicitly references this..

As mention before, 'some sort of user script' does not provide me with any detail/explanation  to go on. If there is a strong reason to include it in the readme, please tell me.

DaveDavenport:
Should we change 'modi' to 'modes' to make it clearer?

(we should update all documentation then to keep it consistent)

DaveDavenport:
changed the modi -> modes.

RayZ0rr:
Hey Dave, I'm sorry for being late. My laptop is at repair shop.

I thought `modi` was your special term for mode. I didn't know you meant plural of mode. That is why all this confusion happened.

Also, won't the commit break existing configurations? Is that necessary? I'm sorry because I am kinda responsible for this. I just wanted to clarify the readme a bit more

DaveDavenport:
No, its backward compatible. The old options are aliased to the new one.
Your confusion was a good indication it was not clear. Hopefully it is better now.

RayZ0rr:
Okay, so they won't be deprecated soon? 

Maybe add a note somewhere on interchangeability between modes and modi? This is so that old users won't be confused when reading upon it in README or man.

Also, close this PR?

Once again thanks for your hardwork.

DaveDavenport:
yeah, `switchers` also still works and that been deprecated ages ago.

PR #1662, opened 2022-07-28
[filebrowser] Fix st_atim undefined on macOS

sainnhe:
When compiling on macOS, an error message will pop up:

```
error: no member named 'st_atim' in 'struct stat'
```

This is because `st_atim`, `st_ctim` and `st_mtim` does not exist on
macOS, instead it uses `st_atimespec`, `st_ctimespec` and
`st_mtimespec`.

In this commit, 3 macros are defined to try to fix this problem.

Reference: https://github.com/hboetes/mg/issues/7

### DIFF
--- a/source/modes/filebrowser.c
+++ b/source/modes/filebrowser.c
@@ -50,6 +50,12 @@
 
 #define FILEBROWSER_CACHE_FILE "rofi3.filebrowsercache"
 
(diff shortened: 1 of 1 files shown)


DaveDavenport:
Did not expect Mac users.
Thanks for the patch.

PR #1685, opened 2022-08-21
Include sys/stat.h for S_IWUSR

jasperla:
Explicitly include `sys/stat.h` to be able to get to `S_IWUSR`. This is required on OpenBSD.


DaveDavenport:
thanks.

PR #1698, opened 2022-09-03
[Themes] Fix and restore Pop-Dark

nadiamoe:
This PR somewhat reverts https://github.com/davatorium/rofi/commit/c0feb2eb0ddfef7f31e179e9585543b1d746cf1c, by restoring `Pop-Dark` and fixing the two missing variables on it.

From the context I understand that the `var(something)` syntax is now preferred to `@something`. If you want, I can probably hack some regex to fix that on the theme too.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -202,6 +202,7 @@ theme_DATA=\
 		   themes/Arc.rasi\
 		   themes/Arc-Dark.rasi\
 		   themes/DarkBlue.rasi\
+		   themes/Pop-Dark.rasi\
 		   themes/Indego.rasi\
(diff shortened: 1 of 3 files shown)


DaveDavenport:
The @ syntax is still fine, the `var()` allows you to set a default.

Maybe you want to update the header of the theme to include yourself  as having  modified it?

nadiamoe:
@DaveDavenport In this case it's a simple two-line change, I don't think it will be necessary :)

DaveDavenport:
Closing as I need correct accreditation for inclusion in distros.

PR #1715, opened 2022-09-26
Add "window-columns" option

kosorin:
This PR add option to remove a pad character for each item in `window-format`.

example

### DIFF
--- a/doc/rofi.1
+++ b/doc/rofi.1
@@ -948,7 +948,7 @@ Format what is being displayed for windows.
 .RE
 
 .PP
-\fIlen\fP: maximum field length (0 for auto-size). If length and window \fIwidth\fP are negative, field length is \fIwidth - len\fP\&.
+\fIlen\fP: maximum field length (0 for auto-size). If length is negative, the entry will be unchanged.
 .br
 If length is positive, the entry will be truncated or padded to fill that length.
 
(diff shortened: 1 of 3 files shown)


DaveDavenport:
Looking at the code, this is not the way to fix it. It adds a config option that breaks the formatting options that already exists.
I think instead of adding a separate option, the existing formatting should be extended: this avoids 'yet another config option', does not break existing formatting options and feels more logical. 

```patch
diff --git a/source/modes/window.c b/source/modes/window.c
index a0e3d136..a6da5b96 100644
--- a/source/modes/window.c
+++ b/source/modes/window.c
@@ -863,6 +863,8 @@ static void helper_eval_add_str(GString *str, const char *input, int l,
   if (l == 0) {
     spaces = MAX(0, max_len - nc);
     g_string_append(str, input_nn);
+  } else if (l < 0) {
+    g_string_append(str, input_nn);
   } else {
     if (nc > l) {
       int bl = g_utf8_offset_to_pointer(input_nn, l) - input_nn;
@@ -891,7 +893,7 @@ static gboolean helper_eval_cb(const GMatchInfo *info, GString *str,
     if (match[2] == ':') {
       l = (int)g_ascii_strtoll(&match[3], NULL, 10);
       if (l < 0) {
-        l = 0;
+        l = -1;
       }
     }
     if (match[1] == 'w') {
```
achieves the same thing without an extra config flag and breaking existing formatting.

kosorin:
Yes this is better. Actually this was first thing I tried before I decided to edit the code, but it didn't work.
So I'll update the PR.

kosorin:
I also changed outdated documentation (1a9dac80).

PR #1735, opened 2022-10-27
input method

duarm:
input method draft, it works, but it was hacked together since I'm not familiar with rofi's code base, nor an ime expert

demo

We just create an XIM handler with a new dependency `xcb-imdkit`, and forward each PRESS event to it. 
only tested with fcitx5.

- [X] fix warning
- [X] window position is slightly off
- [ ] unresponsive typing (because of rofi_set_im_window_pos)
- [ ] first character is delayed


source/xcb.c
```diff
@@ -1550,6 +1561,7 @@ gboolean display_setup(GMainLoop *main_loop, NkBindings *bindings) {
     return FALSE;
   }
   xcb->connection = g_water_xcb_source_get_connection(xcb->source);
+  xcb->im = g_water_xcb_source_get_im(xcb->source);
```
sardemff7: Guess I’ll see a PR on libgwater soon?

source/xcb.c
```diff
@@ -1346,7 +1350,14 @@ static gboolean main_loop_x11_event_handler(xcb_generic_event_t *ev,
     // status);
     return G_SOURCE_CONTINUE;
   }
+
   uint8_t type = ev->response_type & ~0x80;
+  if (!xcb_xim_filter_event(xcb->im, ev) && xcb->ic &&
+      (type == XCB_KEY_PRESS)) {
```
sardemff7: I feel like in some cases, the input-method path would not trigger, and the flat removal of `XCB_KEY_PRESS` handling would be missing  
Also, I didn’t find the doc of `xcb_xim_filter_event()`, but it sounds like you don’t really want it to be hidden in a multiple-conditions thing

I’d rather have:
* the `xcb_xim_filter_event()` call in its own `if` with an early `return` to actually filter xim events **and** with a prefixed `xcb->im != NULL` check, unless you can show me it’ll never be `NULL`, for anyone, ever
* the `xcb->ic != NULL` in the existing `XCB_KEY_PRESS`, to branch the IM vs non-IM behaviours
* the current `XCB_KEY_PRESS` logic in a separate function but **not** the xim callback (to keep only the necessary args in it)

source/view.c
```diff
@@ -799,6 +850,11 @@ void __create_window(MenuFlags menu_flags) {
                        XCB_GRAVITY_STATIC,   XCB_BACKING_STORE_NOT_USEFUL,
                        xcb_event_masks,      map};
 
+  xcb_xim_set_im_callback(xcb->im, &xim_callback, NULL);
+
+  // Open connection to XIM server.
+  xcb_xim_open(xcb->im, open_callback, true, NULL);
```
sardemff7: Check that `xcb->im` is not `NULL` or whatever

source/view.c
```diff
@@ -784,6 +786,55 @@ rofi_view_setup_fake_transparency(widget *win,
...
 }
+
+static void commit_string(xcb_xim_t *im, xcb_xic_t ic, uint32_t flag, char *str,
+                          uint32_t length, uint32_t *keysym, size_t nKeySym,
+                          void *user_data) {
+  (void)user_data;
+  (void)keysym;
+  (void)nKeySym;
+  (void)ic;
+  (void)flag;
```
sardemff7: Use `G_GNUC_UNUSED` in the argument list directly

source/view.c
```diff
@@ -784,6 +786,55 @@ rofi_view_setup_fake_transparency(widget *win,
...
+  if (xcb_xim_get_encoding(im) == XCB_XIM_UTF8_STRING) {
+    fprintf(stderr, "key commit utf8: %.*s\n", length, str);
+  } else if (xcb_xim_get_encoding(im) == XCB_XIM_COMPOUND_TEXT) {
+    size_t newLength = 0;
+    char *utf8 = xcb_compound_text_to_utf8(str, length, &newLength);
+    if (utf8) {
+      int l = newLength;
```
sardemff7: You can drop the debug code

sardemff7 (changes requested):
Not an IM expert, but a few remarks across the board

Looks good from here, with some tweaks it’ll be mergeable in my book

source/xcb.c
duarm: > Also, I didn’t find the doc of xcb_xim_filter_event(), but it sounds like you don’t really want it to be hidden in a multiple-conditions thing

documentation is lacking, but apparently we want to filter all events with xcb_xim_filter, and only forward when them when we are connected with the xim server (xcb->ic). 
https://github.com/fcitx/xcb-imdkit/issues/2#issuecomment-362765322

> xcb->im != NULL check, unless you can show me it’ll never be NULL, for anyone

from my tests, im was never null, I think it's just a container to communicate with the xim server we create on the libgwater part, what defines if there's an IM server or not is by checking if the ic was created (xcb->ic), which I'll fix soon.

but I guess it doesn't hurt to check just to be sure, when I get the time to look through the code, and confirm it, we can remove those checks.

duarm:
related libgwater PR https://github.com/sardemff7/libgwater/pull/4

duarm:
fixed the event forwarding, we only forward the event if we have a xim server in place (when xcb->ic is created), so typing works without an xim server.

added a disconnection callback too, so if the xim server is down, we properly revert to not forwarding events.

both warnings are related to event type 31 (XCB_SELECTION_NOTIFY), for some reason we receive one as soon as we get to the x11 main loop, and rofi returns a `"Failed to convert selection"` if the xim server is down, or a `"Failed"` if the server is up. Not sure why. this isn't affecting anything as far as I know.

XIM server window position is slightly off, the position can be offsetted changing the `open_xim_callback` spot, not sure how to get the necessary information to reposition it in rofi, may be you can help with this one and the warnings?

DaveDavenport:
~~I currently do not keep a relative (or absolute)  x,y position for widgets (that I can remember). This is something I could look at adding. (never really had a use for that information before).~~

Ugh, my brain is foggy. I already do this (for the mouse handling). We just need to recurse back to sum up all relative positions.

Do you need position relative to rofi window or absolute position?

There is a `widget_get_absolute_xpos/ypos`  function you can call on `state->text` that gives you the top left position of the entry box.  Adding `widget_get_height(state->text)`  to y offset should allow you to find the bottom of the widget.

duarm:
> Do you need position relative to rofi window or absolute position?

not sure, I'm passing the CacheState.mainWindow to the server, we define an offset, the window by default (offset 0,0) appears here
image

on my launcher, it appears here
image

>There is a widget_get_absolute_xpos/ypos function you can call on state->text that gives you the top right position of the entry box. Adding widget_get_height(state->text) to y offset should allow you to find the bottom of the widget.

I'll take a look later

sardemff7:
As noted in sardemff7/libgwater#4 the `xcb_xim_t` should be created on the rofi side

source/xcb.c
```diff
@@ -1375,6 +1404,15 @@ static gboolean main_loop_x11_event_handler(xcb_generic_event_t *ev,
   if (xcb->sndisplay != NULL) {
     sn_xcb_display_process_event(xcb->sndisplay, ev);
   }
+
+  if (xcb->im && xcb_xim_filter_event(xcb->im, ev) == 0) {
```
sardemff7: So now, if `xcb_xim_filter_even()` returns `true` (which is, IIUC, “I filtered something, please ignore the event”) you still continue?  
I think it should be
```c
if (xcb->im && xcb_xim_filter_event(xcb->im, ev) ) {
    return G_SOURCE_CONTINUE;
}

/* In the big switch */
    case XCB_KEY_PRESS:
        if ( xcb->ic ) {
            xcb_xim_forward_event(xcb->im, xcb->ic, (xcb_key_press_event_t *)event);
        } else {
            rofi_key_press_event_handler(event, state);
        }
 /* same for KEY_RELEASE */
```

source/xcb.c
```diff
@@ -1375,6 +1404,15 @@ static gboolean main_loop_x11_event_handler(xcb_generic_event_t *ev,
...
+    if (xcb->ic &&
+        (type == XCB_KEY_PRESS || type == XCB_KEY_RELEASE || type == 31)) {
```
sardemff7: See comment above

Also: what is `31` here? Not a great fan of magic values :slightly_smiling_face:

source/xcb.c
```diff
@@ -1375,6 +1404,15 @@ static gboolean main_loop_x11_event_handler(xcb_generic_event_t *ev,
...
+      xcb_xim_forward_event(xcb->im, xcb->ic, (xcb_key_press_event_t *)ev);
```
sardemff7: Is it safe to pass `KEY_RELEASE` events here? (the cast is mostly safe because the structs are built that way, but maybe we shouldn’t still)

DaveDavenport:
I wanted to say:
> * top left..

source/xcb.c
```diff
@@ -1550,6 +1588,7 @@ gboolean display_setup(GMainLoop *main_loop, NkBindings *bindings) {
     return FALSE;
   }
   xcb->connection = g_water_xcb_source_get_connection(xcb->source);
+  xcb->im = g_water_xcb_source_get_im(xcb->source);
```
sardemff7: So, here, I guess
```c
xcb->im = xcb_xim_create(xcb->connection, xcb->screen_nbr, NULL);
```

sardemff7 (changes requested):
We’re getting closer!

source/xcb.c
duarm: i missed this one while I was testing, I was trying to forward the selection event (31) to see if the ime needed that, to no avail, seems to do nothing

duarm:
how can I get the text width? the IM window is already at the correct height for most widgets ([thanks @DaveDavenport), but now I need the current text width so the window follows the text.

DaveDavenport:
> how can I get the text width? the IM window is already at the correct height for most widgets ([thanks @DaveDavenport), but now I need the current text width so the window follows the text.

That is currently not stored.  I will try to add a function that gets the cursor position (I think this is what we want).

duarm:
xim context creation was moved to rofi, a little easier to test now

duarm:
sorry for the next two commits, forgot to unstage the changes

source/xcb.c
duarm: not sure, but the xcb_xim_forward_event only accepts xcb_key_press_event_t, and both implementations I'm using as reference: 
https://github.com/H-M-H/xcb-imdkit-rs/blob/master/src/lib.rs
https://github.com/fcitx/xcb-imdkit/blob/30e2f16f9a8b0e338e25ce5e3643809a07ad41f0/test/client_demo.c

just pass both KEY_PRESS and KEY_RELEASE

DaveDavenport:
I've added a get_cursor_x_pos method to textbox

```
textbox_get_cursor_x_pos(state->text);
```

sardemff7 (commented):
Although it’s some hard-to-read code (as any code related to input, apparently), I’m good with this version on the plumbing side :slightly_smiling_face: 

Guess it just need some polishing on the UI side

I’ll let @DaveDavenport handle the rest :smile: (but don’t hesitate to ping again if needed)

duarm:
tried the new get_cursor_x_pos() and everything's almost working, the problem I found is that get_cursor_x_pos() is returning the cached cursor_pos which is calculated on the previous draw, so just placing the update on rofi_view_handle_text() doesn't work, where should I put the window position update?

Peek 2022-10-27 18-42

duarm:
actually I think the expected behaviour is for the window to follow as you type, before commiting the text. Think I can do that through the callbacks, but that will complicate some things, I would need to preview typed text without actually triggering the fuzzy finder, we might want to do that later, on another PR.

DaveDavenport:
the new position is available once drawn.. So after 'widget_draw'  so in rofi_view_update

duarm:
removed the positioning queue, tested without and had no problems, added a new rofi_set_im_window_pos(x,y)
here's the current behaviour
Peek 2022-10-27 19-05

 the kitty one for reference, we might want to do something like that on the future.
Peek 2022-10-27 18-51

DaveDavenport:
When I run this, I get an unexpected XCB_SELECTION_NOTIFY on startup event. Not sure why.
With no server, its atom 'TRANSPORT'.  I will lower it to a debug msg.
(sorry missed message above).


Do we need to handle this?

DaveDavenport:
What minimal version of imdkit is needed for XCB_XIM_UTF8_STRING?

duarm:
1.0.3, it was added here, https://github.com/fcitx/xcb-imdkit/commit/a16d25ea1ff084c63fea698601687d8142908b2c

duarm:
we can ignore proper encoding handling and just treat all texts as compound on older versions, should I add a macro?

DaveDavenport:
> we ignore proper encoding handling and just treat all texts as compound on older versions, should I add a macro?

if we can support older versions, that should be preferred.  Hopefully fixes ci too

duarm:
the new XCB_IMDKIT_1_0_3_LOWER should be defined by the build system on imdkit lower than 1.0.3, defaulted to using 1.0.3 since ubuntu is already on 1.0.3 apparently
https://launchpad.net/ubuntu/+source/xcb-imdkit

DaveDavenport:
hmm to bad they don't have version info in header.

duarm:
does cmake does that somehow? never used, but they have this, line 4
https://github.com/fcitx/xcb-imdkit/blob/30e2f16f9a8b0e338e25ce5e3643809a07ad41f0/src/CMakeLists.txt

searched through all files tho, nothing, nor any I header I can include on my system

duarm:
after some mundane use, noticed that updating the ime window position (rofi_set_im_window_pos) every frame makes typing unresponsive, think we should only update when committing the text, can we have a `textbox_calculate_cursor_x_pos()`? or you have a better idea?

DaveDavenport:
maybe only send update to the window when position changed?

DaveDavenport:
on older version these:
/home/runner/work/rofi/rofi/builddir/../source/xcb.c:1633: undefined reference to `xcb_xim_set_use_compound_text'
/usr/bin/ld: /home/runner/work/rofi/rofi/builddir/../source/xcb.c:1634: undefined reference to `xcb_xim_set_use_utf8_string
can be disabled?

duarm:
yes, both `xcb_xim_set_use_compound_text` and `xcb_xim_set_use_utf8_string` are from 1.0.3, missed that, sorry

DaveDavenport:
Would this helpw ith the update issue?

```patch
index 58af1cd5..b72fa897 100644
--- a/source/view.c
+++ b/source/view.c
@@ -846,16 +846,18 @@ gboolean rofi_set_im_window_pos(int new_x, int new_y) {
   if (!xcb->ic)
     return false;

-  xcb_point_t spot;
-  spot.x = new_x;
-  spot.y = new_y;
-  xcb_xim_nested_list nested =
-      xcb_xim_create_nested_list(xcb->im, XCB_XIM_XNSpotLocation, &spot, NULL);
-  xcb_xim_set_ic_values(xcb->im, xcb->ic, NULL, NULL, XCB_XIM_XNClientWindow,
-                        &CacheState.main_window, XCB_XIM_XNFocusWindow,
-                        &CacheState.main_window, XCB_XIM_XNPreeditAttributes,
-                        &nested, NULL);
-  free(nested.data);
+  static xcb_point_t spot = {.x = 0, .y = 0};
+  if (spot.x != new_x || spot.y != new_y) {
+    spot.x = new_x;
+    spot.y = new_y;
+    xcb_xim_nested_list nested = xcb_xim_create_nested_list(
+        xcb->im, XCB_XIM_XNSpotLocation, &spot, NULL);
+    xcb_xim_set_ic_values(xcb->im, xcb->ic, NULL, NULL, XCB_XIM_XNClientWindow,
+                          &CacheState.main_window, XCB_XIM_XNFocusWindow,
+                          &CacheState.main_window, XCB_XIM_XNPreeditAttributes,
+                          &nested, NULL);
+    free(nested.data);
+  }
   return true;
 }
 static void open_xim_callback(xcb_xim_t *im, G_GNUC_UNUSED void *user_data) {
```

duarm:
yeah but not by much, its still noticeably slower, I'll hook up with the preedit callback, those are only called when the ime window is open, so if you're typing in english you should pay nothing.

duarm:
hello, I won't be able to work on this for a few days, couldn't make the preedit callbacks work, but aside from this small problem, it's quite ready, feel free to finish or just hack for now. The problem is that if you're typing normal character (not composing a character, like with japanese),  typing really fast is slightly less responsive because of `rofi_set_im_window_pos`.
Once I get back I might try again. Thanks for the good work.

DaveDavenport:
I noticed that with fcitx running, esp the first character takes very long to appear on my test machine. (i5-6500t, 8gb ddr4, nvme disk, not fast, but not a  slow machine either). 
This is kinda not-ok, so might make it a compile time flag.

duarm:
This doesn't happen on my machine, are you on fcitx5 (latest version)? also, when you're using an input method, it sometimes messes your autorepeat (xset r rate) when changing between servers, try running `xset r rate 300 50` or something like that again and see if that was the problem. If it was, we can't do nothing about it.

DaveDavenport:
Installed the fcitx5 (deb) on my pop-os test machine. Its up2date but no idea what version.
But there was a clear 'delay' on the first character on first run.

DaveDavenport:
@sardemff7  is that 'changes requested' still open or can this be merged?

sardemff7 (approved):

DaveDavenport:
Users report (and shown) extreme slowdown by this patch making rofi near unusable.
(see the discussion I tagged @duarm  in)

I will revert unless we can find a solution.

duarm:
revert this for now, tested yersterday under heavy load and it's quite unusuable indeed. There's something wrong with the event forwarding, sometime the keys are swalloed without the string being commited, only to both be commited once the next event is forwarded.  This becomes apparent if the code cant run fast enough, which is caused by the window position being updated every draw as I've noted before, we're paying to much round-trip with the server.

Ideally this shouldn't happen, even under heavy loads, this pr requires some more careful handling with the server (only communicate when necessary, throttling, etc.), I'll try to fix when I get the time again, or someone with more knowledge can step in.

Issue #1217, opened 2020-11-08
Set ellipsize mode in config file?


areographe:
Hey, I noticed that this (or the similar feature of setting ellipsize mode in the command line flags) has been requested [here](https://www.reddit.com/r/qtools/comments/e1v0jd/is_there_a_way_to_start_rofi_in_right/) and [here](https://www.reddit.com/r/qtools/comments/f0o62g/default_ellipsize_option/) earlier this year.

I couldn't see an open feature request on github for it so I'm making one now :)

# Use case

I have a lot of PDFs open when I'm doing academic work, with really long window titles due to the PDF reader showing the full path (a Zotero library with a lot of subfolders), so I generally have to toggle the ellipsis placement (to middle of the line) with <kbd>Alt</kbd><kbd>.</kbd> every time I switch windows with rofi. Would be great to be able to specify this default in `config.rasi`.

Issue #399, opened 2016-05-08
Allow for left/right scroll


sprig:
Hello,

It would be nice if, when run in 'dmenu' mode, one could scroll the results of long strings right and left, and if it was possible to select right/left justification of the list of results.

Would you care to comment on where in the source it would be proper to add this functionality?

DaveDavenport:
right/left align is done based if the text is RTL it is right align LTR 
This will right align:
```
echo -e "\u200f aap" | rofi -dmenu
```

The scrolling is probably not that trivial.  Scrolling text on hover/selected row might be useful.

sprig:
Thanks for the tip!

Scrolling selected row would be useful, but I would rather dynamic R/L justification, if it were a choice.

Usecase:

I have written a small custom script to search files based on filename, in my homedir, for which the UI was via dmenu and was bound to hotkey (e.g. Caps+b then, type a keyword from a paper name). This works surprisingly well, and much faster than anything normal file managers give, for my usecase. The full paths can be quite long (e.g. take almost entire screen width in small font). In rofi, most of the path is obstructed by '...', and esp. the most important part, which is the actual filename.

I got around this by piping to rofi just the filename and making rofi output just line number, which is used to select the line of actual output with the full filename. It would be better, however, if I could examine the entire path because sometimes I have similarly named files residing in different places.

Scroll on select would help here, but I suspect that it will be painfully slow to sift through the files this way.

DaveDavenport:
If I get it right, you want to select between ellipsizing method at runtime.. (select between end, beginning or middle)

sprig:
This is not precisely what I want, but it would be a major step forward in functionality for my use case ideally I would still like to be able to see entire paths somehow, but I understand that since every path has a different length, it is not obvious what is the correct way to implement scrolling. 

Perhaps dynamic ellipsizing together with showing the entire text after a (selectable at invocation) amount of of time for selected text would be 95%+ of what I want.

DaveDavenport:
Having all the text 'scroll'  all the time is not a good solution. This will be very annoying to watch (and hard to read), I have tried this in another program (gmpc's playlist) in the past: It does not work.

Not sure what the right solution is here.

sprig:
I agree - I didn't want all the text autoscrolling all the time. I wanted functioning left/right keys :-)

DaveDavenport:
aah I see.  (Those are already mapped for navigating the text box entry.)

do you want to 'scroll' all entries at once?

[...]

PR #1747, opened 2022-11-19
[WIP] Ellipsize modes

cognitiond:
This PR creates additional functions for setting ellipsize mode; to middle and end along with the already existing start function, and adds command line options to control the initial ellipsize mode: `-ellipsize-mode (start|middle|end)`.

Some things though:
* Is this the right place to implement this?
* How should I give an error? Right now, I'm following how -scroll-method does it, and just reverting to default if invalid, and printing out an error message to stderr, but I'm not sure if this is the right way to print out an error message, especially since I had to forward-declare the error printing function.
* Is the option (-ellipsize-mode) too long?
* This (currently) doesn't apply to the dmenu mode, should I duplicate this code into dmenu_mode_dialog() or is there a better way.

Thanks :)

### DIFF
--- a/config/config.c
+++ b/config/config.c
@@ -106,6 +106,7 @@ Settings config = {
     .tokenize = TRUE,
     .matching = "normal",
     .matching_method = MM_NORMAL,
+    .ellipsize_mode = "end",
 
     /** Desktop entries to match in drun */
     .drun_match_fields = "name,generic,exec,categories,keywords",
--- a/include/settings.h
+++ b/include/settings.h
@@ -98,6 +98,8 @@ typedef struct {
   SortingMethod sorting_method_enum;
   /** Sorting method. */
   char *sorting_method;
+  /** How to ellipsize an entry */
+  char *ellipsize_mode;
 
   /** Desktop entries to match in drun */
   char *drun_match_fields;
--- a/include/view.h
+++ b/include/view.h
@@ -338,9 +338,21 @@ void rofi_capture_screenshot(void);
 void rofi_view_set_window_title(const char *title);
 
 /**
- * set ellipsize mode to start.
+ * Set ellipsize mode to start.
  */
 void rofi_view_ellipsize_start(RofiViewState *state);
+
+/**
+ * Set ellipsize mode to middle.
+ */
+void rofi_view_ellipsize_middle(RofiViewState *state);
+/** @} */
+
+/**
+ * Set ellipsize mode to end.
+ */
+void rofi_view_ellipsize_end(RofiViewState *state);
+/** @} */
 
 /**
  * @param new_x New XIM window x pos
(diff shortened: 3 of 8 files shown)


DaveDavenport:
What is the use-case for this addition?

cognitiond:
You know how people can read words even when the middle letters are mixed up? Having the start and end of a string is much more useful than having only either end, for example with filesystem paths, or URLS, or command lines, or really just any long string.
Also, it was requested by #1217 and [this comment](https://github.com/davatorium/rofi/issues/399#issuecomment-614819946).

More generally, it's just exposing underlying functionality that already exists, but can't currently be accessed by the user.

DaveDavenport:
> You know how people can read words even when the middle letters are mixed up? 

I have no clue how typoglycemia is relevant.  Do you want an option that scrambles the  characters in the middle of certain words?

In general the requirement for this to work for people, from what I've read in the past (it does no work for me unless I typed it myself, thx dyslexia ), is that the word length does not change and the scrambled word is relatively short. No words are removed and function words are not scrambled, etc..  This is not something ellipsizing does, it is even very unlikely as you use ellipsizing to shorten the whole line...


I was (partially) asking for the use-case to see , if we add this option, how to best add this option..
We could also make a flag on listview to set default ellipsize mode.

```css
listview {
ellipsize: middle;
}
```
or on the actual textbox? 
```css
element-text {
ellipsize: middle;
}
```
2ndly  I am partially unsure for the use-case for this.  Why would you want to display content that you already know is going to be to long to display most of the time (for the rare to long string, we have the keybinding)?
In case of filepaths+names,  why show the full path name if you then set an option to cut most of it off again?
Wouldn't it be better to format it so its more useful, f.e.:
```<b>{filename}</b> <span size='x-small'>{directory}</span>```


For quick'n'dirty script I can see the use for this, so adding this to dmenu kinda makes more sense to me.

cognitiond:
Being able to set the initial ellipsize mode has been requested by people for years:
* https://github.com/davatorium/rofi/issues/1217
* https://www.reddit.com/r/qtools/comments/f0o62g/default_ellipsize_option/
* https://www.reddit.com/r/qtools/comments/e1v0jd/is_there_a_way_to_start_rofi_in_right/
* https://github.com/davatorium/rofi/issues/399#issuecomment-614819946

If you think that it fits best in the dmenu mode, well you probably know better than I do, but I don't see why not allow users to choose the ellipsize mode for any mode, since you can already toggle ellipsizing in any mode. The functionality to set the ellipsis mode already exists in rofi, so why not allow them to choose it themselves, if their use case requires it?

DaveDavenport:
Why is wanting a valid use-case before adding a feature wrong?

Every option needs to be maintained, as you saw with the recent 'keep-right' issue, even simple options can break.  I don't have a lot of time to work on rofi, I rather spend it on options that have a use or I enjoy figuring out and adding.  Instead I am having to deal with misinformation about typoglycemia, my questions being ignored and having to repeat myself.

I've added he option in dmenu, as I can see a use-case there. I have not added it to the normal mode as I don't see the use-case (as explained); I don't want 'promote' custom mode scripts that rely on modifying global options to work properly instead of working well with default settings.

Issue #1751, opened 2022-11-20
[REQUEST] Launch programs in dedicated cgroups


mweinelt:
### Before opening a feature request

- [X] I checked the *next* branch to see if the feature has already been implemented
- [X] I searched existing reports to see if it is already requested.

### What is the user problem or growth opportunity you want to see solved?

I want better out of memory management than what the Linux kernel provides. So I went and enabled `systemd-oomd.service` on my machine. But it cannot work properly because everything rofi starts is launched in the `session.scope`, not a dedicated cgroup.

> Be aware that if you intend to enable monitoring and actions on user.slice, user-$UID.slice, or their ancestor cgroups, it is highly recommended that your programs be managed by the systemd user manager to prevent running too many processes under the same session scope (and thus avoid a situation where memory intensive tasks trigger systemd-oomd to kill everything under the cgroup). If you're using a desktop environment like GNOME or KDE, it already spawns many session components with the systemd user manager. 

https://www.freedesktop.org/software/systemd/man/systemd-oomd.service.html#System%20requirements%20and%20configuration

### How do you know that this problem exists today? Why is this important?

Looking through `systemd-cgtop` reveals many cgroups in the system.slice (per systemd unit), a few in the user.slice (user units), and a large session-1.scope, where every app launched through rofi lands in.

### Who will benefit from it?

Users on distros that run with systemd by default can enable oomd to prevent stalls due to out of memory usage. Oomd will have the opportunity to kill the correct cgroup instead of the coarse `session-1.scope`.

### Rofi version (rofi -v)

1.7.5+wayland1

### Configuration

n/a

DaveDavenport:
Can't this be achieved by changing the run*command options?

Also if rofi is launched in the user cgroup, should the programs it launch also wind up in the user cgroup?  If I launch firefox from rofi on my system, it winds up in user.slice.

mweinelt:
Indeed, that would be possibly, let me look into that.

DaveDavenport:
I am not cgroup expert (used it in the past, but never at 'desktop level') but shouldn't it inherit from its parent, so as long as rofi runs in the right slice, so will the applications it launches.
Given rofi is launched by the user in the user session, it should run in the user slice to begin with.

mweinelt:
The GNOME people decoupled programs started from the launcher and through Alt+F2 from the launchers scope in this PR:  https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/863/commits

[...]

PR #1752, opened 2022-11-20
[DOC] Add example run command with cgroup support

mweinelt:
Adds an example command to launch processes via systemd-run into a dedicated systemd scope within the users slice, which allows using systemd-oomd, that evaluates memory consumption on a per cgroup basis.

Related to #1751

### DIFF
--- a/doc/rofi.1.markdown
+++ b/doc/rofi.1.markdown
@@ -543,6 +543,12 @@ See *PATTERN*.
 
 Default: *{cmd}*
 
+Example to run applications in a dedicated cgroup with systemd. Requires a shell to escape and interpolate the unit name correctly.
+
+```
+"bash -c 'systemd-run --user --unit=app-rofi-\$(systemd-escape {cmd})-\$RANDOM {cmd}'"
+```
+
 `-run-shell-command` *cmd*
 
 Set command to execute when running an application in a shell.
--- a/doc/rofi.1
+++ b/doc/rofi.1
@@ -902,6 +902,18 @@ See \fIPATTERN\fP\&.
 Default: \fI{cmd}\fP
 
 .PP
+Example to run applications in a dedicated cgroup with systemd. Requires a shell to escape and interpolate the unit name correctly.
+
+.PP
+.RS
+
+.nf
+"bash -c 'systemd-run --user --unit=app-rofi-\\$(systemd-escape {cmd})-\\$RANDOM {cmd}'"
+
+.fi
+.RE
+
+.PP
 \fB\fC-run-shell-command\fR \fIcmd\fP
 
 .PP


doc/rofi.1.markdown
```diff
@@ -543,6 +543,12 @@ See *PATTERN*.
 
 Default: *{cmd}*
 
+Example to run applications in a dedicated cgroup with systemd:
+
+```
+systemd-run --user --scope --unit='{cmd}-$(date +%s)' {cmd}
```
mweinelt: The hierarchy looks like this with the given command:

````
├─user.slice
  └─user-1000.slice
    ├─user@1000.service
      ├─session.slice
        ├─app.slice
          └─firefox-1668960502.scope
````

The timestamp is useful to give each scope a unique and meaningful name.

**Comment by mweinelt** [comment]
Reading up on https://systemd.io/DESKTOP_ENVIRONMENTS/.

**Comment by mweinelt** [comment]
````
├─app-rofi-foot-30370.service
````

Unfortunately we can't embed the PID into the unit name, so I went with `$RANDOM` instead, to get unique names without requiring the subshell as proposed in the initial version.

**Comment by DaveDavenport** [comment]
$random does not work when set in config file for me:

```
Invalid unit name "app-rofi-/usr/bin/google-chrome-stable-$RANDOM" escaped as "app-rofi--usr-bin-google-chrome-stable-\x24RANDOM" (maybe you should use systemd-escape?).
```

**Comment by DaveDavenport** [comment]
ugly but works-ish:
```
bash -c 'systemd-run --user --unit=app-rofi-{cmd}-$RANDOM {cmd}'
```

**Comment by DaveDavenport** [comment]
ugly but works-ish:
```
bash -c 'systemd-run --user --unit=app-rofi-{cmd}-$RANDOM {cmd}'
```

**Comment by DaveDavenport** [comment]
ugly but works-ish:
```
bash -c 'systemd-run --user --unit=app-rofi-{cmd}-$RANDOM {cmd}'
```

**Comment by DaveDavenport** [comment]
wth did github do..

**Comment by mweinelt** [comment]
> rofi -show drun -show-icons -run-command "systemd-run --user --unit='app-rofi-{cmd}-$RANDOM' -p CollectMode=inactive-or-failed {cmd}"

This is what I'm running right now. I think this is about the quoting.

````
$ rofi -show drun -show-icons -run-command "systemd-run --user --unit='app-rofi-{cmd}-$RANDOM' -p CollectMode=inactive-or-failed {cmd}"


(process:2735559): GLib-CRITICAL **: 19:57:55.137: g_string_insert_len: assertion 'len == 0 || val != NULL' failed
Running as unit: app-rofi-gedit-12124.service

**Comment by mweinelt** [comment]
> Invalid unit name "app-rofi-/usr/bin/google-chrome-stable-$RANDOM" escaped as "app-rofi--usr-bin-google-chrome-stable-\x24RANDOM" (maybe you should use systemd-escape?).

Dashes need to be escaped, that's why the name is invalid.

Maybe it should be

````
"systemd-run --user --unit='app-rofi-$(systemd-escape {cmd})-$RANDOM' {cmd}"
````

The double quotes on the outside are essential for `$RANDOM` to be interpolated.

**Comment by DaveDavenport** [comment]
rofi does not use a shell to execute the command, so $() won't work. unless you use a shell. Same for `$RANDOM`

**Comment by mweinelt** [comment]
It does work like this from within sway with exec:

````
bindsym Mod4+d exec /nix/store/c4bl07a1fnbnimvcmbsa0hsd4pfy11a7-rofi-1.7.5+wayland1/bin/rofi -show drun -show-icons -run-command "systemd-run --user --unit='app-rofi-{cmd}-$RANDOM' -p CollectMode=inactive-or-failed {cmd}"

**Comment by DaveDavenport** [comment]
because it is expanded when sway calls rofi via a shell.

**Comment by mweinelt** [comment]
We do need something unique at the end, so we don't reference the same cgroup over and over. I've looked through the available patterns, but none seemed useful for that purpose.

**Comment by DaveDavenport** [comment]
```
rofi -show drun -show-icons -run-command "systemd-run --user --unit='app-rofi-{cmd}-$RANDOM' -p CollectMode=inactive-or-failed {cmd}" -h | grep run
	-run-command [string]                   Run command to execute
		systemd-run --user --unit='app-rofi-{cmd}-16317' -p CollectMode=inactive-or-failed {cmd} (Commandline)
```

Not sure ,  maybe launching via shell is the best, it should be short lived, I assume that systemd-run will not block?

**Comment by mweinelt** [comment]
Won't unless you ask it to `--wait`.

**Comment by mweinelt** [comment]
> maybe launching via shell is the best

Updated the documentation to say as much.

**Comment by DaveDavenport** [comment]
I don't think this is very clear for users, its also incomplete. $RANDOM requires a shell that supports it.
Maybe better to provide an example that just works (both in config as cmdline):

```
configuration {
	run-command: "bash -c 'systemd-run --user --unit=app-rofi-$(systemd-escape {cmd})-$RANDOM {cmd}'";
}
```
That also works when passed on commandline:

```bash
rofi -show drun -run-command "bash -c 'systemd-run --user --unit=app-rofi-$(systemd-escape {cmd})-$RANDOM {cmd}'"
```

mweinelt:
Ah thanks, I didn't consider the configuration file!

Unfortunately the systemd-escaping happens before {cmd} is substituted?!

```
Running as unit: app-rofi-x7bcmdx7d-7635.service

**Comment by DaveDavenport** [comment]
aah, that probably needs an extra escape on commandline.

**Comment by DaveDavenport** [comment]
```bash
rofi -show drun -run-command "bash -c 'systemd-run --user --unit=app-rofi-\$(systemd-escape {cmd})-\$RANDOM {cmd}'"
```

mweinelt:
That looks good from where I stand! Thanks for sticking with me on this!

DaveDavenport:
I think we might need to update this:
https://systemd.io/DESKTOP_ENVIRONMENTS/

```
rofi -show drun -run-command "bash -c 'systemd-run  --user --unit=app-org.rofi.\$(systemd-escape {cmd})@\$RANDOM {cmd}'"
```
?

mweinelt:
Ah, I went off of this remark.

> TODO: Define the name of slices that should be used. This could be `app-<launcher>-<ApplicationID>-<RANDOM>.slice`.

I've just checked a machine with GNOME 42, and it has e.g. `app-gnome-firefox-2948.scope` , `app-gnome-org.gnome.SettingsDaemon.DiskUtilityNotify-1810.scope` or `app-gnome-writer-6818.scope`.

Launcher and Application Id do need to be separated by a dash.

Issue #788, opened 2018-03-31
Decrease cursor width?


navdeepio:
Is it possible?

DaveDavenport:
No.

Please follow the [guidelines](https://github.com/DaveDavenport/rofi/blob/next/.github/CONTRIBUTING.md) when creating an issue.

navdeepio:
@DaveDavenport Please consider this issue as a feature request

DaveDavenport:
@navdeepio Please read my comment from 31st of March.

PR #1753, opened 2022-11-23
make cursor more customizable by adding cursor-width and cursor-color

vE5li:
This pull request aims to make cursor more customizable by adding `cursor-width` and `cursor-color` to the theme. This will allow the creation of block cursors such as this:
block_cursor
and also address issue #788.

### DIFF
--- a/doc/rofi-theme.5
+++ b/doc/rofi-theme.5
@@ -197,6 +197,25 @@ element-text {
 .RE
 
 .PP
+We can also specify the color and width of the cursor. You could, for example,
+create a crimson block cursor like this:
+
+.PP
+.RS
+
+.nf
+entry {
+  cursor-color: rgb(220,20,60);
+  cursor-width: 8px;
+}
+
+.fi
+.RE
+
+.PP
+By default, the \fB\fCcursor-color\fR will be the same as the \fB\fCtext-color\fR\&. The \fB\fCcursor-width\fR will always default to 2 pixels.
+
+.PP
 If you want to see the complete theme, including the modification you can run:
 
 .PP
--- a/doc/rofi-theme.5
+++ b/doc/rofi-theme.5
@@ -1532,6 +1532,10 @@ This option is only available on the \fB\fCelement-text\fR widget.
 Set the location of tab stops by their distance from the beginning of the line.
 Each distance should be greater than the previous one.
 The text appears to the right of the tab stop position (other alignments are not supported yet).
+.IP \(bu 2
+\fBcursor-width\fP:      The width of the cursor.
+.IP \(bu 2
+\fBcursor-color\fP:      The color used to draw the cursor.
 
 .RE
 
--- a/doc/rofi-theme.5.markdown
+++ b/doc/rofi-theme.5.markdown
@@ -132,6 +132,18 @@ element-text {
 └─────────────────────────────────────────────────────────────────────┘ 
 ```
 
+We can also specify the color and width of the cursor. You could, for example,
+create a crimson block cursor like this:
+
+```css
+entry {
+  cursor-color: rgb(220,20,60);
+  cursor-width: 8px;
+}
+```
+
+By default, the `cursor-color` will be the same as the `text-color`. The `cursor-width` will always default to 2 pixels.
+
 If you want to see the complete theme, including the modification you can run:
 
 ```bash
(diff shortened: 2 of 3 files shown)


DaveDavenport:
looking at the code, it looks (on first sight) like this breaks placeholder-color.

DaveDavenport:
I compiled and checked, it breaks the place-holder color.

DaveDavenport:
Not very awake, but I think this fixes things:

```patch
diff --git a/source/widgets/textbox.c b/source/widgets/textbox.c
index bda7d6fe..1838ccff 100644
--- a/source/widgets/textbox.c
+++ b/source/widgets/textbox.c
@@ -482,9 +482,12 @@ static void textbox_draw(widget *wid, cairo_t *draw) {
   }
   y += top;
 
-  if (tb->show_placeholder) {
-    rofi_theme_get_color(WIDGET(tb), "placeholder-color", draw);
-  }
+  // TODO check if this is still needed after flatning.
+  cairo_set_operator(draw, CAIRO_OPERATOR_OVER);
+  cairo_set_source_rgb ( draw, 0.0, 0.0, 0.0 );
+  // use text color as fallback for themes that don't specify the cursor color
+  rofi_theme_get_color(WIDGET(tb), "text-color", draw);
+
   // Set ARGB
   // We need to set over, otherwise subpixel hinting wont work.
   switch (pango_layout_get_alignment(tb->layout)) {
@@ -515,11 +518,6 @@ static void textbox_draw(widget *wid, cairo_t *draw) {
   }
   }
 
-  // TODO check if this is still needed after flatning.
-  cairo_set_operator(draw, CAIRO_OPERATOR_OVER);
-  cairo_set_source_rgb ( draw, 0.0, 0.0, 0.0 );
-  // use text color as fallback for themes that don't specify the cursor color
-  rofi_theme_get_color(WIDGET(tb), "text-color", draw);
 
   // draw the cursor
   if (tb->flags & TB_EDITABLE) {
@@ -553,6 +551,9 @@ static void textbox_draw(widget *wid, cairo_t *draw) {
   // draw the text
   cairo_save(draw);
   cairo_reset_clip(draw);
+  if (tb->show_placeholder) {
+    rofi_theme_get_color(WIDGET(tb), "placeholder-color", draw);
+  }
   pango_cairo_show_layout(draw, tb->layout);
   cairo_restore(draw);
 }
```

this also fixes the cursor taking the place-holder color.

vE5li:
You are completely right. I applied your suggestions and tested again, now everything seems to work as expected.

DaveDavenport:
Can you add the properties to the textbox in the rofi-themes manpage?

vE5li:
Sure thing! I will find time for it this evening

vE5li:
I was struggling a bit to come up with some good documentation and what I ended up with is definitely not perfect but hopefully good enough. Please feel free to modify it if there's something missing or you don't like the phrasing :)

DaveDavenport:
It might be good to also add it to the textbox section?

```patch
diff --git a/doc/rofi-theme.5.markdown b/doc/rofi-theme.5.markdown
index bf0f3630..93da4bd0 100644
--- a/doc/rofi-theme.5.markdown
+++ b/doc/rofi-theme.5.markdown
@@ -937,6 +937,8 @@ The following properties are currently supported:
     Set the location of tab stops by their distance from the beginning of the line.
     Each distance should be greater than the previous one.
     The text appears to the right of the tab stop position (other alignments are not supported yet).
+* **cursor-width**:     The width of the cursor.
+* **cursor-color**:      The color used to draw the cursor.
 
 ### listview:
 * **columns**:         integer
```

vE5li:
Yes definitely, added them just now.

I noticed that the capitalization on that section is quite inconsistent. For example:
```
* **font**:              the font used by this textbox (string).
* **str**/**content**:   the string to display by this textbox (string).
* **vertical-align**:    Vertical alignment of the text. A number between 0 (top) and 1 (bottom).
* **horizontal-align**:  Horizontal alignment of the text. A number between 0 (left) and 1 (right).
* **text-color**:        the text color to use.
* **text-transform**:    text style {color} for the whole text.
* **highlight**:         text style {color}.
```
But I wasn't sure if there is maybe some rule for when to capitalize that I'm missing. I would love to fix this in a follow up MR if it makes sense.

DaveDavenport:
Documentation MR's are highly appreciated.

vE5li:
Alright, I will look into that :) Is this MR ready to be merged?

PR #1765, opened 2022-12-20
[WIP] Improve require-input user experience

elamperti:
Whenever the listview has `require-input` set to true, if the user hasn't entered any text, pressing kb-accept will result in the selection of the first item of the list, even if it's not visible for the user.  Also trying to navigate the list (blindly) doesn't change the selected item either.

My proposed change prevents auto-selecting the first option blindly, avoiding an unexpected default selection (whatever is the first item in the listview). With this change, rofi remains open until the user either enters the _required input_ and accepts, or presses kb-cancel. This goes in line with the UX of other search/filtering tools, and it should benefit the community users' who use this option.

See more details and ongoing discussion [here](https://github.com/davatorium/rofi/discussions/1763).

### DIFF
--- a/source/view.c
+++ b/source/view.c
@@ -1717,7 +1717,7 @@ static void rofi_view_trigger_global_action(KeyBindingAction action) {
   }
   case ACCEPT_ENTRY: {
     rofi_view_refilter_force(state);
-    // If a valid item is selected, return that..
+    // If a valid item is selected, return that.
     unsigned int selected = listview_get_selected(state->list_view);
     state->selected_line = UINT32_MAX;
     if (selected < state->filtered_lines) {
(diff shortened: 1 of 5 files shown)


source/widgets/listview.c
```diff
@@ -577,6 +577,9 @@ void listview_set_num_elements(listview *lv, unsigned int rows) {
 }
 
 unsigned int listview_get_selected(listview *lv) {
+  if (lv->require_input && !lv->filtered) {
+    return UINT32_MAX;
+  }
```
elamperti, line 582: This is enough to avoid auto-selecting the first item when `require_input` is enabled and there's no input:

source/view.c
```diff
@@ -1728,7 +1728,9 @@ static void rofi_view_trigger_global_action(KeyBindingAction action) {
       state->retv = MENU_CUSTOM_INPUT;
     }
 
-    state->quit = TRUE;
+    if (selected != UINT32_MAX) {
+      state->quit = TRUE;
+    }
```
elamperti, line 1733: And checking the selected value here, before defining it's time to quit, would keep rofi open when `require_input` is enabled and there's no input. The user can still quit rofi normally by using the expected method, kb-cancel.

PR #1790, opened 2023-01-23
[rofi-theme-selector] Replace `$HOME` with a tilde

yusufaktepe:
This change will make it easier to share the configuration with different users.


yusufaktepe:
@DaveDavenport I noticed that in https://github.com/davatorium/rofi/commit/09e0f876581ff6388af1faf8ce5b100511744693, you preferred appending over changing the line. Is there a reason for this?

DaveDavenport:
because sed modifying a @theme is not a valid way to completely change to a new theme.

This should be obvious:

```css

@theme "mytheme"

@import "mytheme-mods"

window {
   fullscreen: true;
}
```

PR #1791, opened 2023-01-25
[rofi-theme-selector] prepend newline before specifying new theme

amithm7:
If the EOF is not a newline, new theme setting will fail.

### DIFF
--- a/script/rofi-theme-selector
+++ b/script/rofi-theme-selector
@@ -200,7 +200,7 @@ set_theme()
     fi
 
     # Comment old base theme, not replace as it may be modified after the line
-    ${SED} -i 's/^\s*\(@theme\s\+".*"\)/\/\/\1/' "${get_link}"
+    ${SED} -i 's,^\s*\(@theme\s\+".*"\),\/\/\1,' "${get_link}"
     echo -e "\n@theme \"${1}\"" >> "${get_link}"
 }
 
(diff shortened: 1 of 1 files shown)


script/rofi-theme-selector
```diff
@@ -198,8 +198,10 @@ set_theme()
     then
        touch "${get_link}"
     fi
-    ${SED} -i 's/^\s*\(@theme\s\+".*"\)/\/\/\1/' "${get_link}"
-    echo "@theme \"${1}\"" >> "${get_link}"
+
+    # Comment old base theme, not replace as it may be modified after the line
+    ${SED} -i 's,^\s*\(@theme\s\+".*"\),\/\/\1,' "${get_link}"
```
sardemff7: Since you changed the delimiter from `/` to `,` you don’t need to escape the `/` in the replacement

Also, a nicer way to do that is to use line matching and “replace” the beginning of the line:
```suggestion
    ${SED} -i '/^\s*@theme\s\+"/s:^://:' "${get_link}"
```
I don’t think you need to match the full line with both double quotes or whatever, since it is a rofi syntax error  
I would even argue you can just match `^@theme` without anything else, so something like that:
```suggestion
    ${SED} -i '/^\s*@theme/s:^://:' "${get_link}"
```

amithm7: Great :+1:

PR #1792, opened 2023-01-25
Add support for adding textbox widgets to listview elements

notuxic:
This PR adds support for adding custom _textbox_ widgets as children of _element_ widgets.

This allows themes to add static text to entries that can be styled according to state (selected, alternate, ...).

My personal use-case is a powerline-theme, that uses custom _textbox_ widgets for the powerline-symbols.
Before, only custom _box_ widgets were supported (ie. no support for displaying text), and using _*-display-format_ does not allow for different styling depending on state.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -29,6 +29,7 @@
 #define ROFI_VIEW_H
 
 #include "mode.h"
+#include "widgets/widget.h"
 #include <pango/pango.h>
 #include <xcb/xcb.h>
 /**
@@ -354,6 +355,11 @@ void rofi_view_ellipsize_listview(RofiViewState *state,
  * main_window
  */
 gboolean rofi_set_im_window_pos(int new_x, int new_y);
+
+
+WidgetTriggerActionResult textbox_button_trigger_action(
+    widget *wid, MouseBindingMouseDefaultAction action, G_GNUC_UNUSED gint x,
+    G_GNUC_UNUSED gint y, G_GNUC_UNUSED void *user_data);
 
 /** @} */
 #endif
--- a/source/view.c
+++ b/source/view.c
@@ -2115,7 +2115,7 @@ static int rofi_view_calculate_height(RofiViewState *state) {
   return widget_get_desired_height(main_window, state->width);
 }
 
-static WidgetTriggerActionResult textbox_button_trigger_action(
+WidgetTriggerActionResult textbox_button_trigger_action(
     widget *wid, MouseBindingMouseDefaultAction action, G_GNUC_UNUSED gint x,
     G_GNUC_UNUSED gint y, G_GNUC_UNUSED void *user_data) {
   RofiViewState *state = (RofiViewState *)user_data;
--- a/source/widgets/listview.c
+++ b/source/widgets/listview.c
@@ -36,6 +36,7 @@
 
 #include "settings.h"
 #include "theme.h"
+#include "view.h"
 
 #include "timings.h"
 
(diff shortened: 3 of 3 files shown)


DaveDavenport:
I am unsure what this achieves?  you add an textbox widget to the listview that can only have a pre-defined text set in the theme?

notuxic:
As mentioned above, I personally use it to add powerline-symbols to the entries. They are drawing-characters that usually use the background-color of the preceding text as foreground-color to archive the "powerline-look". Since I have different background-colors depending on state, I also need different foreground-colors for the powerline-symbols. That is no possible using pango-markup.
Other people might find other uses as well.

Also, adding custom textboxes is supported in other places as well (for example in the inputbar), so not being able to add a textbox to listview elements was unexpected behaviour (I spent some time debugging why the text of my "textbox" wasn't appearing before figuring out that _rofi_ just created a regular _box_, even though my widget was named _textbox*_ as required by the docs for adding custom widgets)

DaveDavenport:
I try to not make assumptions when reading requests (to avoid making wrong ones) and to me it was not completely clear what you where trying to do.  I *think* i get it now?  People reading this might not have all context you assume known.
(a screenshot would have been very clarifying.) 

Maybe we should also allow adding a settable icon widget?

notuxic:
>  I think i get it now? People reading this might not have all context you assume known.
(a screenshot would have been very clarifying.)

Here's a screenshot:

Capture

I set the foreground-color of the powerline-symbol to red on the alternate lines, to make it clearer. It's first a `element-icon`, then a custom textbox with the triangle-symbol, then the `element-text`. By setting the right foreground- and background-colors for the symbols you get this "powerline-look".

> Maybe we should also allow adding a settable icon widget?

IMO, the listview elements should support all custom widgets for consistency, I just didn't want to add it to the initial PR right away. I can add it now though.

DaveDavenport:
I think making it consistent is a good idea. Initially listview only allowed a certain set of subwidgets (as also mentioned in the manpage) 
```
   • element: a box in the listview holding the entries
     • element-icon: the widget in the listview's entry showing the (optional) icon
     • element-index: the widget in the listview's entry keybindable index (1,2,3..0)
     • element-text: the widget in the listview's entry showing the text.
````

Supporting boxes was something I did testing out something (if you have multiple columns to simulate a centered view when only one column worth of elements where left). 
I agree making it consistent is good thing to do.

notuxic:
I've now added support for adding custom icons and buttons as well (not sure what buttons could be useful for, maybe `accept-alt`/`accept-custom`).

I've also looked through the documentation if there's anything conflicting with adding custom widgets to listview elements, but found nothing that had to be changed (the part you mentioned above is about the _default_ hierachy, which also mentions `inputbar` which _can_ already have custom children).

I had to remove the `static` from `textbox_button_trigger_action` and add a declaration to the header file, since that function is also needed in `listview.c` now. Not sure if you'd want to "refactor" that some other way.

DaveDavenport:
Agree that documentation is not the clearest, things grown over time.
Improvement would be welcome. 

the children of the input bar can be moved outside of the inputbar, the children of the listview and element cannot. We should explain this better.

I'll try to clarify this: https://github.com/davatorium/rofi/blob/6caece77d4cc0fa81b3303a4126e4efbb93b41be/doc/rofi-theme.5.markdown?plain=1#L991 a bit. This then needs to be updated for your change.

DaveDavenport:
If documentation is updated (I tried to clarify the current situation  a bit), ill merge this.

PR #1795, opened 2023-01-29
[Script] Strip pango markup when matching rows

tpope:
This is effectively the same fix that was made to DMenu in a42e9f869f8fe95772aace1544dcf907827a0e52.


DaveDavenport:
Thanks.

Issue #1806, opened 2023-02-11
[REQUEST] Pre-check some or all boxes on `-demu -multi-select`


KSXGitHub:
### Before opening a feature request

- [X] I checked the *next* branch to see if the feature has already been implemented
- [X] I searched existing reports to see if it is already requested.

### What is the user problem or growth opportunity you want to see solved?

When I launch `rofi -dmenu -multi-select`, all boxes are unchecked by default, but my use case requires all boxes are checked by default, and user may manually uncheck undesired boxes.

### How do you know that this problem exists today? Why is this important?

Rofi dmenu use case is similar to a terminal based fuzzy finder program, both capable of multi-selection, except Rofi dmenu can't pre-check boxes.

### Who will benefit from it?

People who want to use rofi as a fuzzy finder

### Rofi version (rofi -v)

Version: 1.7.5

### Configuration

https://gist.github.com/KSXGitHub/40e74f10d3aec68d92c748333bb11bab

DaveDavenport:
Please test PR.

DaveDavenport:
no longer interrested in implementing this.  A good PR is welcome.

PR #1807, opened 2023-02-11
[Dmenu] Implement a select rows for dmenu multi-select.

DaveDavenport:
Issue: #1806

>
> - [ ] This PR targets the `next` branch and not `master`
> - [ ] Its commits' summaries are reasonably descriptive
> - [ ] You've described what this PR addresses below
> - [ ] You've included links to relevant issues, if any with `#issue_num`
>

Your description here...

### DIFF
--- a/doc/rofi-dmenu.5
+++ b/doc/rofi-dmenu.5
@@ -148,6 +148,13 @@ Or any combination: '5,-3:,7:11,2,0,-9'
 Urgent row, mark \fIX\fP as urgent. See \fB\fC-a\fR option for details.
 
 .PP
+\fB\fC-select-rows\fR \fIX\fP
+
+.PP
+If multi-select is enabled, pre-select rows, See \fB\fC-a\fR option for format details.
+If same row is specified multiple times, it state is toggled on subsequential sets.
+
+.PP
 \fB\fC-only-match\fR
 
 .PP
--- a/doc/rofi-dmenu.5.markdown
+++ b/doc/rofi-dmenu.5.markdown
@@ -93,6 +93,11 @@ Active row, mark *X* as active. Where *X* is a comma-separated list of python(1)
 
 Urgent row, mark *X* as urgent. See `-a` option for details.
 
+`-select-rows` *X*
+
+If multi-select is enabled, pre-select rows, See `-a` option for format details.
+If same row is specified multiple times, it state is toggled on subsequential sets.
+
 `-only-match`
 
 Only return a selected item, do not allow custom entry.
--- a/doc/rofi-theme.5
+++ b/doc/rofi-theme.5
@@ -1615,7 +1615,7 @@ property.
 These cannot be changed using the \fB\fCchildren\fR property.
 
 .PP
-Each entries displayed by listview are captured by a \fB\fCbox\fR called \fB\fCelement\fR\&.
+Each Entry displayed by listview is captured by a \fB\fCbox\fR called \fB\fCelement\fR\&.
 An \fB\fCelement\fR widget can contain the following special child widgets:
 
 .RS
@@ -1630,7 +1630,8 @@ An \fB\fCelement\fR widget can contain the following special child widgets:
 
 .PP
 By default the \fB\fCelement-icon\fR and \fB\fCelement-text\fR child widgets are added to the
-\fB\fCelement\fR\&. This can be modified using the \fB\fCchildren\fR property.
+\fB\fCelement\fR\&. This can be modified using the \fB\fCchildren\fR property or the
+\fB\fC[no]-show-icons\fR option.
 
 .PP
 A child added with another name is seen as a \fB\fCbox\fR, this can be used as dynamic
--- a/source/modes/dmenu.c
+++ b/source/modes/dmenu.c
@@ -123,6 +123,66 @@ typedef struct {
   DmenuModePrivateData *pd;
 } Block;
 
+
+static void dmenu_parse_multi_select_range ( DmenuModePrivateData *pd, const char *entries)
+{
+  if ( entries == NULL ) {
+    return;
+  }
+  // Pre-alloc array.
+  if (pd->selected_list == NULL) {
+    pd->selected_list =
+      g_malloc0(sizeof(uint32_t) * (pd->cmd_list_length / 32 + 1));
+  }
+  char *entries_cp = g_strdup ( entries );
+  char *endp;
+  const char *const sep = ",";
+  for (char *token = strtok_r(entries_cp, sep, &endp); token != NULL;
+    token = strtok_r(NULL, sep, &endp)) {
+      const char *sep[] = {"-", ":"};
+      int pythonic = (strchr(token, ':') || token[0] == '-') ? 1 : 0;
+      int index = 0;
+
+      int start = -1;
+      int stop = -1;
+      for (char *inner_token = strsep(&token, sep[pythonic]); inner_token != NULL;
+        inner_token = strsep(&token, sep[pythonic])) {
+          if (index == 0) {
+            start = stop = (int)strtol(inner_token, NULL, 10);
+            index++;
+            continue;
+          }
+
+          if (inner_token[0] == '\0') {
+            stop = -1;
+            continue;
+          }
+
+          stop = (int)strtol(inner_token, NULL, 10);
+          if (pythonic) {
+            --stop;
+          }
+        }
+        // Fix negative numbers.
+        if ( start < 0 ) {
+          start = pd->cmd_list_length + start;
+        }
+        if ( stop < 0 ) {
+          stop = pd->cmd_list_length + stop;
+        }
+        // Fix starting
+        for ( int index = start; index <= stop; index++ ){
+          if ( index < 0 ) {
+            index = pd->cmd_list_length - index;
+          }
+          if ( index < (int)pd->cmd_list_length ) {
+            bittoggle(pd->selected_list, index);
+          }
+        }
+    }
+  g_free ( entries_cp );
+}
+
 static void read_add_block(DmenuModePrivateData *pd, Block **block, char *data,
                            gsize len) {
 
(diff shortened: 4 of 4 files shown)


doc/rofi-dmenu.5
```diff
@@ -147,6 +147,13 @@ Or any combination: '5,-3:,7:11,2,0,-9'
 .PP
 Urgent row, mark \fIX\fP as urgent. See \fB\fC-a\fR option for details.
 
+.PP
+\fB\fC-select-rows\fR \fIX\fP
+
+.PP
+If multi-select is enabled, pre-select rows, See \fB\fC-a\fR option for format details.
+If same row is specified multiple times, it state is toggled on subsequential sets.
```
KSXGitHub, line 155: I think this behavior is surprising. I prefer the box is checked if the row id appears more than once, and unchecked if the row id doesn't appear.

Is there any reason you think toggling based on odd/even number of occurence a superior behavior?

doc/rofi-dmenu.5
```diff
@@ -147,6 +147,13 @@ Or any combination: '5,-3:,7:11,2,0,-9'
 .PP
 Urgent row, mark \fIX\fP as urgent. See \fB\fC-a\fR option for details.
 
+.PP
+\fB\fC-select-rows\fR \fIX\fP
```
KSXGitHub, line 151: The term "select" can be confusing since it already used to mean highlighting the current item in the list before either confirm selection or check the box.

Then again, Rofi has already name the flag `-multi-select` so it could be that the maintainers do care about this term.

doc/rofi-dmenu.5
DaveDavenport: It would make it so much easier to select all rows except  a number of rows...
I am trying to make things easier for the user.

doc/rofi-dmenu.5
DaveDavenport: It is the term that is used everywhere else.. Using another term here would be confusing.
Multi-select is a nice to have and does not completely fit anyway.

doc/rofi-dmenu.5
KSXGitHub: I see.

I would have suggested adding an option named `-unselect-rows`, but then I realized it cannot substitute `-select-rows $FOO,$BAR,$BAZ` to mean selecting `$BAZ` and `$FOO` except `$BAR`.

So now the question is: This python-like range syntax is also used in other options as well, do these options also support this toggling behavior? If they do, it is consistent. If they don't, it means that the range syntax is lacking, and instead of adding toggling behavior, we should improve the range syntax (in other PR/discussion ofc).

KSXGitHub:
I just cloned this branch and tested it. It work great so far.

doc/rofi-dmenu.5
KSXGitHub: > This python-like range syntax is also used in other options as well, do these options also support this toggling behavior?

There is another option called `-a` to make "active rows". And it **does not have** the toggling behavior. Which means that `-a 1,2,3,2` would make 1, 2, 3 active whilst `-select-rows 1,2,3,2` only checks 1 and 3.

DaveDavenport: Neither '-a' or '-u' document that it has the toggling behaviour. I have no idea why you think it does?
I expect that developers of script that use advanced features like this will read the documentation and can understand the difference.
Its a feature I can support for select-row, but not (easily) for the others, I can see this as a benefit for the script developer.

KSXGitHub: If you say so. I only use rofi for a few use cases, so this quirk won't matter to me that much right now.

If I have sowed doubt in you, then you should ask other developers and users to see what they think. If you are confident that this is best approach, just merge it.

DaveDavenport: ??? 

You make a statement **in bold** that says -a does not support it. Given it nowhere claims it does or even suggests this,  I am confused by this remark. That is all I pointed out.

_I don't understand the 'sowed doubt in you'  remark.._ 


If anybody is wiling to update the `-u`,`-a` to support toggling that would be appreciated.  It is going to a bit more work to support this (and work correctly with dmenu async mode.).
I rather not add code to remove functionality, but add/rework code to lift all the functionality equal.

KSXGitHub: > If anybody is wiling to update the `-u`,`-a` to support toggling that would be appreciated.

Ah. I see what you mean. Although, even if it can be achieved, suddenly add toggling to an existing option that didn't support it before could be a breaking change.

It is true that if the documentation doesn't state that it supports a feature, then it won't. And users should read documentation. However, documentation doesn't make inconsistency disappears from an interface. Take JavaScript for example, `[] + {}` being `"[object Object]"` and `{} + []` being 0 is well documented, but still considered unexpected and inconsistent.

I appreciate that you think ahead and add more features to make the tool more versatile. However, (mis)features are easier added than removed. This is a new feature, it is an opportunity avoid footguns.

How about this? Introduce a new syntax that is superset of the range syntax: `-select-rows 'range1 & !(range2 & !range3)'` should select (`range1` excluding `range2`) and `range3`. Alternatively, `-select-rows 'range1 \ (range2 \ range3)'`. Toggling behavior is of course not needed.

DaveDavenport: > Ah. I see what you mean. Although, even if it can be achieved, suddenly add toggling to an existing option that didn't support it before could be a breaking change.

~This option did not exist before, `-select-rows` was added in this PR.
No existing option has been changed, no backwards compatibility with existing options has been broken in this PR.~ misread this, my bad.

if you mean it if we add this later, it would be a breaking change yes.
I don't know if people really do specify overlapping regions already. That might indeed be a dangerous assumption.

>How about this? Introduce a new syntax that is superset of the range syntax: -select-rows range1 & !range2 should select range1 excluding range2. Alternatively, -select-rows range1 \ range2. Toggling behavior is of course not needed.

I don't see how this is more consistent. But more important, this  will introduce a shell escaping nightmare as &  and ! have special meaning .

KSXGitHub: > No existing option has been changed, no backwards compatibility with existing options has been broken in this PR.

I was talking about "If anybody is wiling to update the -u,-a to support toggling that would be appreciated."

> this will introduce a shell escaping nightmare as & and ! have special meaning

There is always other syntax beside `&`, `!`, and `\`. The range syntax doesn't support letters, right? so the range expression (let's call it that) can always add a few keywords. For example, `-select-rows 'range1 excluding (range2 excluding range3)'` or `-select-rows 'range1 and not(range2 and not range3)'`.

> I don't see how this is more consistent

Because it does not make the range syntax of `-select-rows` to have special meaning of toggling behavior despite other similar options not supporting it. Also, if someone wishes to introduce the range expression syntax to `-a` and `-u` in the future, it would not be a breaking change.

DaveDavenport: (yes sorry, I misread and corrected my post already. It seems we did this in parallel).

I was saying about the consistency because the proposed solution  ('&','!' ) or the new one now ( 'and' and 'not') is not the type of syntax we in rofi.
If this is so important we should not introduce another one.

To specify multiple items, we use ',' separator in several places. 

```css 
entry, prompt {
}
window {
children: [inputbar, listview];
}
```
or 
```bash
rofi -show drun -modes 'drun,window'
```

So if we list multiple ranges we should stick to ',' as list separator.

To negate we use '-' in the search entry box. But this is not ideal for this usecase.
(you can search for 'browser -chrome'  to search all non chrome browser.)

KSXGitHub: > is not the type of syntax we in rofi.

You need to complete this sentence or I wouldn't understand.

> If this is so important we should not introduce another one.
> 
> To specify multiple items, we use ',' separator in several places.

Reusing `,` wouldn't make sense because `,` is for adding items, not for intersecting set of items. In other words, `,` is an `or` operator, different from an `and` operator. `0:4,2:5` is `0,1,2,3,4` whereas `0:4 and 2:5` would be `2,3`.

> If this is so important

It's not important for my use cases right now, neither is the toggling feature, I only brought this up because I dislike the toggling behavior.

KSXGitHub: > To negate we use '-' in the search entry box. But this is not ideal for this usecase.
> (you can search for 'browser -chrome' to search all non chrome browser.)

Because negative numbers are a thing.

This gives me an idea: Many websites have a search syntax similar to GitHub issue (`k1:v1 k2:v2 pattern`). So rofi could potentially introduce a similar syntax: `0:100 not:(10:90 not:40:50)`; or, if you want the `not` keyword to be more easily discernible from normal text: `0:100 @not:(10:90 @not:40:50)`.

DaveDavenport: >>    is not the type of syntax we in rofi.

> You need to complete this sentence or I wouldn't understand.

  is not the type of syntax we use in rofi.

     

> Reusing , wouldn't make sense because , is for adding items, not for intersecting set of items. In other words, , is an or operator, different from and and operator. 0:4,2:5 is 0,1,2,3,4 whereas 0:4 and 2:5 would be 2,3.

eeuh what?? I was talking about adding items not intersecting,  so ',' makes sense for having a list of ranges that should be selected. (As it is now). We use ',' to separate in a list already. How we specify what we then exclude then is not clear.

If we start describing thing in set theory..  then we should use the right syntax for that.. I would be fine with using  ∪,∩, \  (union, intersection, set diff if i remember correctly)  But I think we will have lost most of the developers/users by now.


Anyway I give up. If anybody wants to pick this up and create a PR , I'll reconsider.
I spend more time then I care to spend on/have for this already

KSXGitHub: OK. I have created https://github.com/davatorium/rofi/pull/1809 which removes the toggling behavior.

I only proposed a variety of different syntaxes to replace toggling because I dislike it.

Issue #1817, opened 2023-02-27
[BUG] element does not support textbox children


friedow:
### Rofi version (rofi -v)

1.7.5

### Configuration

https://gist.github.com/friedow/b7f5504d73fb24a5debf469009429e12

### Theme

https://gist.github.com/friedow/b7f5504d73fb24a5debf469009429e12

### Launch command

rofi -show drun

### Step to reproduce

1. Create a theme containing
``` rasi
* {
    // font: "monospace 12";
    background-color: black;
    text-color: white;
}

element {
    children: [ element-text, textbox-foo ];
}

textbox-foo {
    content: "foo";
    background-color: green;
}
```
1. Launch rofi with `rofi -show drun`
1. This should produce a rofi screen with green boxes without content:
![image](https://user-images.githubusercontent.com/17351844/221599174-38822303-2827-4a2d-bb48-355245546d5c.png)

### Expected behavior

The green boxes should contain the string `foo`.

### Actual behavior

The green boxes contain nothing.

### Additional information

I would love to fix that myself if you dont mind. I think I already found the correct place to fix this. You can expect a pull request for this in the next few days :).

### Using wayland display server protocol

- [X] No, I don't use the wayland display server protocol

DaveDavenport:
https://github.com/davatorium/rofi/pull/1792

This fixes adding any widget to the listview, it just waits for a documentation update (to explain the change it makes).

DaveDavenport:
btw your example already works, except you have no space assigned to textbox-foo

```css
* {
    // font: "monospace 12";
    background-color: black;
    text-color: white;
}

element {
    children: [ element-text, textbox-foo ];
}

textbox-foo {
    content: "foo";
    background-color: green;
    expand: true;
}
```

DaveDavenport:
![rofi-2023-02-27-1622-00000](https://user-images.githubusercontent.com/406245/221604619-550b686e-317f-4ab0-b73f-724b61f0b396.png)

DaveDavenport:
~ok~ sorry, I got confused by
> This should produce a rofi screen with green boxes without content:

It does if you fix the theme you pasted.

The content should be fixed by the above PR, if you can add the missing documentation I can merge it. Otherwise it will be merged once I have time to add documentation.

PR #1818, opened 2023-02-28
docs: element children theming

friedow:
This adds documentation for theming of elements. This is the documentation for https://github.com/davatorium/rofi/pull/1792. This is part of https://github.com/davatorium/rofi/issues/1817.

### DIFF
--- a/doc/rofi-theme.5.markdown
+++ b/doc/rofi-theme.5.markdown
@@ -1011,7 +1011,7 @@ By default the `element-icon` and `element-text` child widgets are added to the
 `[no]-show-icons` option.
 
 A child added with another name is treated the same as the special widget described
-in the [advanced layout](https://github.com/davatorium/rofi/blob/next/doc/rofi-theme.5.markdown#advanced-layout) section.
+in the [advanced layout](#advanced-layout) section.
 
 #### listview text highlight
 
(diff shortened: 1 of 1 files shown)


DaveDavenport:
I think this part of the documentation is a better place to modify it:

> A child added with another name is seen as a box, this can be used as dynamic padding to tweak how an entry looks when it expands to available space.

(just before: https://github.com/davatorium/rofi/blob/next/doc/rofi-theme.5.markdown#listview-text-highlight) 

Maybe something like:

A child added with another name is treated the same as the special widget described in the advanced layout section.

(where 'advanced layout' is a link to the advanced layout section anchor).

https://github.com/davatorium/rofi/blob/next/doc/rofi-theme.5.markdown#advanced-layout

friedow:
Good point, this location is much more suitable! I've patched that.

doc/rofi-theme.5.markdown
```diff
@@ -1010,8 +1010,8 @@ By default the `element-icon` and `element-text` child widgets are added to the
 `element`. This can be modified using the `children` property or the
 `[no]-show-icons` option.
 
-A child added with another name is seen as a `box`, this can be used as dynamic
-padding to tweak how an entry looks when it expands to available space.
+A child added with another name is treated the same as the special widget described
+in the [advanced layout](https://github.com/davatorium/rofi/blob/next/doc/rofi-theme.5.markdown#advanced-layout) section.
```
DaveDavenport: Maybe not an absolute link, but relative in current document?

> [advanced layout](#advanced-layout)
```
[advanced layout](#advanced-layout)
```

friedow: Done. I didn't know if the manpage generation could handle relative links ;).

PR #1827, opened 2023-03-29
[WIP] scroll-multiplier option

djvs:
This PR adds a "scroll-multiplier" option to help with paginating tons of content.  I'm guesstimating how some of this stuff works so could use a review.  Docs also not added (not positive which ones to put it in).


DaveDavenport:
Can you explain what this PR does/adds?

djvs:
If you use the scroll wheel to navigate through a long list of results, it'll go 8 lines with a scroll-multiplier of 8, instead of 1, for each tick of the scroll wheel.

DaveDavenport:
Doesn't this patch also change the behaviour of the keybindings?

To me it feels weird to just skip pages. I am not sure this is desirable?  I would more assume a 'kinetic?' scrolling, so it speeds up when your keep scrolling.  If you skip 8 pages, it might be hard to get to the item on page 4 now?

djvs:
Oh, yeah, it does, I can fix the binding issue.    But yeah it only goes a fraction of a page at a time with a reasonably small scroll multiplier.

DaveDavenport:
In keyb.c the scroll action is bound to a certain 'action'. (see rofi -show keys for the bindings).

You could f.e. bind this to page up/down if you want to jump bigger, or we should make a new action that does (kinetic) scrolling.

djvs:
I pushed a fix for that - it's not "kinetic" as in with a decaying trajectory over time and all that jazz, but you could extend it like that in the future.

PR #1838, opened 2023-04-27
Add wezterm to rofi-sensible-terminal

yutkat:
I would like to add wezterm as it is a popular terminal these days and has many users.

https://github.com/wez/wezterm

PR #1840, opened 2023-05-03
Update rofi-sensible-terminal to include wezterm

lokesh-krishna:
Adds wezterm so that users using wezterm, a fairly popular terminal, launching .desktop files with `terminal = true` don't have to deal with the "Failed to find suitable terminal" message.


DaveDavenport:
duplicate #1838 the previous PR?

PR #1901, opened 2023-09-27
Fix text color when `cursor-color` is set

vE5li:
There was a regression in commit a5bd8bc63095d80ea3ad6f462138c23dc57c4e74 that makes text render in the same color as the cursor when `cursor-color` is set. This fixes the issue by correctly saving the state before setting the cursor color and restoring it afterwards.


DaveDavenport:
thanks.

PR #1939, opened 2024-01-28
[WIP] xdg thumbnails fetching with fallback on mimetype icons

giomatfois62:
This implements xdg thumbnails fetching functionality in rofi-icon-fetcher.  

When an icon is requested for an entry in filebrowser mode, and its name is prefixed by the string "thumbnail://", an existing thumbnail is searched in `$HOME/.cache/thumbnails/` and loaded as icon for the entry.   

If the thumbnail does not exists, an icon for the entry mime-type is loaded instead.  

If the entry is a ".desktop" file, its icon key is used instead to load the icon.  

WIP because it lacks the functionality to generate missing thumbnails using system thumbnailers and an option for users to use their own scripts as custom thumbnailers.


DaveDavenport:
Thanks, looks good.. What is the license of the md5 code? can we include it?

giomatfois62:
it seems very much free to use. from the [LICENSE](https://github.com/Zunawe/md5-c/blob/main/UNLICENSE)  

> This is free and unencumbered software released into the public domain.
> Anyone is free to copy, modify, publish, use, compile, sell, or
> distribute this software, either in source code form or as a compiled
> binary, for any purpose, commercial or non-commercial, and by any
> means.

Don't know if you prefer it as a 3rd party submodule, last commit was almost a year ago

DaveDavenport:
Great. We should copy the license into the header of the file then, so it is clear how it is licensed. 

Some things I noted trying it out:

- [ ] Fix meson build
- [ ] Should we be more agrassive in finding thumbnail? for me it fails as it looks in wrong directory.

I will try it out further.

Thanks!

DaveDavenport:
rofi-2024-01-28-1825-00000

Seems to work!

giomatfois62:
Sorry I completely forgot about meson, I tend to use configure.  
As for the missing thumbnails, i look for them in the folder returned by `g_get_user_cache_dir`, on my system it points to $HOME/.cache

DaveDavenport:
That is not what I tried to point out.. for me it looks for thumbnail that does exists, but only larger.

I worked around this as a test like this:

```patch
 // build file thumbnail's path using md5 hash of its encoded uri string
-static gchar* rofi_icon_fetcher_get_file_thumbnail(const gchar* file_path, 
-                                                   int requested_size, 
+static gchar *rofi_icon_fetcher_get_file_thumbnail(const gchar *file_path,
+                                                   int requested_size,
                                                    int *thumb_size) {
   // convert filename to encoded uri string and calc its md5 hash
   gchar *encoded_entry_name = g_filename_to_uri(file_path, NULL, NULL);
@@ -319,49 +319,71 @@ static gchar* rofi_icon_fetcher_get_file_thumbnail(const gchar* file_path,
   md5_to_hex(md5_hex, hex_size, md5, md5_size);
 
   // determine thumbnail folder based on the request size
-  const gchar* cache_dir = g_get_user_cache_dir();
-  gchar* thumb_path;
+  const gchar *cache_dir = g_get_user_cache_dir();
+  gchar *thumb_path = NULL;
 
   if (requested_size <= 128) {
     *thumb_size = 128;
-    thumb_path = g_strconcat(cache_dir, "/thumbnails/normal/",
-        md5_hex, ".png", NULL);
-  } else if (requested_size <= 256) {
+    thumb_path =
+        g_strconcat(cache_dir, "/thumbnails/normal/", md5_hex, ".png", NULL);
+  }
+  if (thumb_path && g_file_test(thumb_path, G_FILE_TEST_EXISTS)) {
+    return thumb_path;
+  }
+  g_free(thumb_path);
+  thumb_path = NULL;
+
+  if (requested_size <= 256) {
     *thumb_size = 256;
-    thumb_path = g_strconcat(cache_dir, "/thumbnails/large/",
-        md5_hex, ".png", NULL);
-  } else if (requested_size <= 512) {
+    thumb_path =
+        g_strconcat(cache_dir, "/thumbnails/large/", md5_hex, ".png", NULL);
+  }
+
+  if (thumb_path && g_file_test(thumb_path, G_FILE_TEST_EXISTS)) {
+    return thumb_path;
+  }
+  g_free(thumb_path);
+  thumb_path = NULL;
+
+  if (requested_size <= 512) {
     *thumb_size = 512;
-    thumb_path = g_strconcat(cache_dir, "/thumbnails/x-large/",
-        md5_hex, ".png", NULL);
-  } else {
+    thumb_path =
+        g_strconcat(cache_dir, "/thumbnails/x-large/", md5_hex, ".png", NULL);
+  }
+  if (thumb_path && g_file_test(thumb_path, G_FILE_TEST_EXISTS)) {
+    return thumb_path;
+  }
+  g_free(thumb_path);
+  thumb_path = NULL;
+
+  {
     *thumb_size = 1024;
-    thumb_path = g_strconcat(cache_dir, "/thumbnails/xx-large/",
-        md5_hex, ".png", NULL);
+    thumb_path =
+        g_strconcat(cache_dir, "/thumbnails/xx-large/", md5_hex, ".png", NULL);
   }
 
   return thumb_path;
 }
```

I did something like this to make it find more thumbnails.

DaveDavenport:
This is not ideal,  I think we should reverse it.. find the biggest one exists (we will auto-scale it down) and work our way down..

Because now I don't see it in a theme with bigger 'icons'. (I only have thumbs in normal folder).

Is there a nice way to trigger the generation of thumbnails? are these script shipped?

giomatfois62:
> Is there a nice way to trigger the generation of thumbnails? are these script shipped?

That's the next step, it's still a work in progress.  There is a specification for thumbnailers that is used by file managers. They are ".desktop" files placed in /usr/share/thumbnailers" and "$HOME/.local/share/thumbnailers", containing a list of compatible mime-types and the corresponding to command to generate their thumbnails.  
The command takes some arguments coded as follows:  
- %i is replaced by the filename
- %u is replaced by an encoded uri of the filename
- %o is the output file path of the thumbnail
- %s is the requested size (max between width and height)  

I'm looking at the implementation in [libgnome-desktop](https://gitlab.gnome.org/GNOME/gnome-desktop/-/blob/master/libgnome-desktop/gnome-desktop-thumbnail.c?ref_type=heads) and was thinking of using a similar solution.  

Create an hashmap of mimetypes as keys and compatible thumbnailers as values. This can be placed in the global state of rofi-icon-fetcher-data, and populated at creation reading from the thumbnailers standard directories.  
When a thumbnail is not found in the cache, look its mimetype in the hashmap and call the corresponding thumbnailer command expanding the arguments, we have already filename, encoded uri and target size.  

Doing this will make thumbnails generated by rofi compatible with other file managers

giomatfois62:
Seems to work pretty good now, I suggest you to first install the [xapp-thumbnailers](https://github.com/linuxmint/xapp-thumbnailers), this way you have thumbnailers available for epub, mp3, raw files and others.  
rofi-2024-01-30-2101-00000
rofi-2024-01-30-2105-00000
rofi-2024-01-30-2106-00000
rofi-2024-01-30-2110-00000
rofi-2024-01-30-2112-00000

DaveDavenport:
that looks very nice.

giomatfois62:
to complete this work, i would like to add the ability to specify a custom thumbnailer command when launching rofi.  
I was thinking something like a "-preview-cmd" or "-icon-fetcher" key that takes a string with the command or path to the script to call. If that is specified, we can call it instead of the standard thumbnailers for mimetypes.  

We could keep saving the thumbnails in the xdg cache directories, the custom script would take as arguments the entry name, the output path of the thumbnail (generated with the md5 hash of the entry name) and an integer size. Then, in the script one could do cool things based on the size requested, like for a list of games fetching their icons for small previews and high res banners for bigger previews.  

A practical example would be something like this:

`rofi -show-icons -preview-cmd "/path/to/custom/script %i %o %s"`  

%i, %o and %s will then be substituted by entry name, ouptut image path and requested size in rofi-icon-fetcher, the code to do that is already there to use standard thumbnailers.  

What do you think?

DaveDavenport:
sounds good..  maybe look at available helper functions.. There is already code to launch applications and do substitution. 
(different format, using {} like {name} )

https://github.com/davatorium/rofi/blob/next/source/modes/ssh.c#L97 example ( need to check, this leaks memory in argsv?)

giomatfois62:
You mean helper_parse_setup and helper_execute?  
I looked at helper_execute before, but we need to wait for the thumbnailer processes to finish, so I used g_spawn_sync instead.

DaveDavenport:
helper_parse_setup mostly, as it does the replacement.

giomatfois62:
> helper_parse_setup mostly, as it does the replacement.

I see, it does exactly what I need. Thanks for the pointer!

giomatfois62:
I tried using helper_parse_setup to build to list of arguments of the thumbnailers, but I'm facing an issue. I need to replace the strings "%i", "%o", "%s" but the function is leaving them unchanged.
This is my call:  
`helper_parse_setup(command, &command_args, &argsv, "{%i}", filename,
      "%u", encoded_uri, "%o", output_path, "%s", size_str, (char *)0);`  
I tried using some encodings like "%%i", "\%i", "\\%i" and others to no avail.

DaveDavenport:
It does not support replacing '%i'.. it replaces anything between brackets..

so `ssh {host}`  it replaces `{host}` with the actual hostname. 
In the above example I would make the keywords '{filename}' '{uri}' '{output}'  or something similar.

giomatfois62:
okay thanks, now i understand, we can use it for the custom thumbnailer scripts but not for the xdg thumbnailers installed in the system.  
Then I'll use it to better conform with the rest of rofi custom command options.  

I was testing performance using the new recursive browser mode, and I got a little worried about performance, cpu is trashed and scrolling becomes very very slow after some pages.  
I want to check if there is some way to set process niceness to make the thumbnailers more resource friendly.

UPDATE: solved prepending "nice -n 19" to the command arguments list, much much better performance

DaveDavenport:
In theory it should be 'async' from the UI, but trashing your I/O it often becomes more difficult for the scheduler to do a good job.. 

I hope to be able to play with the PR this weekend.

Thanks.

giomatfois62:
I pushed the latest changes, it now works for me also using a custom script. To test, run something like this:  
`echo ... | rofi -dmenu -show-icons -preview-cmd "/path/to/custom/script {input} {output} {size}"`  
and put something in the echo to set the icon, like  
`"A\x00icon\x1fthumbnail://A\nB\x00icon\x1fthumbnail://B\nC\x00icon\x1fthumbnail://C"`

DaveDavenport:
took me a while to get it working as on debian apparmour was blocking evince-thumbnailer from writing files when launched  by rofi.. 
(we should include this in documentation)..

It works really well, I will play with it some more later..

I did notice that for some reason fallback icon did not work anymore, but did not get around debugging this.

giomatfois62:
> took me a while to get it working as on debian apparmour was blocking evince-thumbnailer from writing files when launched by rofi.. (we should include this in documentation)..

I had some problems too, some commands were not recognized despite being in /usr/bin, I had to set it as the working directory of the spawned process to make it work. That's not very clean as a workaround to be honest.  

> I did notice that for some reason fallback icon did not work anymore, but did not get around debugging this.  

That's strange, i see them in filebrowser mode, there was something broken with the last "custom command" addition but it should be fixed with the last commit, I did not touch anything about mimetype icons fallback.

DaveDavenport:
> ../source/rofi-icon-fetcher.c:167:3: error: unknown type name ‘GStrvBuilder’
  167 |   GStrvBuilder *cmd_builder = g_strv_builder_new();
  
  I think we need to bump the glib dependency to use this method.

https://docs.gtk.org/glib/ctor.StrvBuilder.new.html

2.68 so that excludes ubuntu 20.04 LTS

giomatfois62:
> 2.68 so that excludes ubuntu 20.04 LTS  
that's unfortunate, I'll try to use a different method. Building an array of strings shouldn't be too hard

source/rofi-icon-fetcher.c
```diff
@@ -91,6 +100,150 @@ typedef struct {
...
+    for (i = 0; command_parts[i] != NULL; i++) {
+      if (strcmp(command_parts[i], "%i") == 0) {
+        g_strv_builder_add(cmd_builder, filename);
+      } else if (strcmp(command_parts[i], "%u") == 0) {
+        g_strv_builder_add(cmd_builder, encoded_uri);
+      } else if (strcmp(command_parts[i], "%o") == 0) {
+        g_strv_builder_add(cmd_builder, output_path);
+      } else if (strcmp(command_parts[i], "%s") == 0) {
+        char size_str[33];
+        sprintf(size_str, "%d", size);
```
DaveDavenport: To stop static code analysis tools complaining, lets use snprintf.

DaveDavenport:
There is a similar version bump required for 
https://docs.gtk.org/glib/func.spawn_check_wait_status.html

2.70.

not sure what minimum distro version this results in, but might be worth it bumping up?

source/rofi-icon-fetcher.c
giomatfois62: ok this one was easy :)

DaveDavenport:
quick hack to make it compile on this machine (still has some debug).
I did something wrong, because the icons are generated, but not loaded (on next rofi load I see them).
```patch
diff --git a/source/rofi-icon-fetcher.c b/source/rofi-icon-fetcher.c
index 533acb84..b6ce3a6b 100644
--- a/source/rofi-icon-fetcher.c
+++ b/source/rofi-icon-fetcher.c
@@ -38,6 +38,7 @@
 #include "settings.h"
 #include <cairo.h>
 #include <pango/pangocairo.h>
+#include <sys/wait.h>
 
 #include "keyb.h"
 #include "view.h"
@@ -164,38 +165,47 @@ static gchar** setup_thumbnailer_command(const gchar *command,
   gchar **command_parts = g_strsplit(command, " ", 0);
   gchar **command_args = NULL;
   
-  GStrvBuilder *cmd_builder = g_strv_builder_new();
-  
   if (command_parts) {
+    int index = 0;
+    command_args = g_malloc0(4*sizeof(char*));
     // set process niceness value to 19 (low priority)
-    g_strv_builder_add(cmd_builder, "nice");
-    g_strv_builder_add(cmd_builder, "-n");
-    g_strv_builder_add(cmd_builder, "19");
-    
+    command_args[index++]= g_strdup("nice");
+    command_args[index++]= g_strdup("-n");
+    command_args[index++]= g_strdup("19");
+    //
     // add executable and arguments of the thumbnailer to the list
     guint i;
     for (i = 0; command_parts[i] != NULL; i++) {
       if (strcmp(command_parts[i], "%i") == 0) {
-        g_strv_builder_add(cmd_builder, filename);
+        command_args = g_realloc(command_args, (index+2)*sizeof(char*));
+        command_args[index++] = g_strdup(filename);
+        command_args[index] = NULL;
       } else if (strcmp(command_parts[i], "%u") == 0) {
-        g_strv_builder_add(cmd_builder, encoded_uri);
+        command_args = g_realloc(command_args, (index+2)*sizeof(char*));
+        command_args[index++] = g_strdup(encoded_uri);
+        command_args[index] = NULL;
       } else if (strcmp(command_parts[i], "%o") == 0) {
-        g_strv_builder_add(cmd_builder, output_path);
+        command_args = g_realloc(command_args, (index+2)*sizeof(char*));
+        command_args[index++] = g_strdup(output_path);
+        command_args[index] = NULL;
       } else if (strcmp(command_parts[i], "%s") == 0) {
-        char size_str[33];
-        sprintf(size_str, "%d", size);
-        g_strv_builder_add(cmd_builder, size_str);
+        command_args = g_realloc(command_args, (index+2)*sizeof(char*));
+        command_args[index++] = g_strdup_printf("%d", size);
+        command_args[index] = NULL;
       } else {
-        g_strv_builder_add(cmd_builder, command_parts[i]);
+        command_args = g_realloc(command_args, (index+2)*sizeof(char*));
+        command_args[index++] = g_strdup(command_parts[i]); 
+        command_args[index] = NULL;
       }
     }
 
-    command_args = g_strv_builder_end(cmd_builder);
+    for ( int i =0; i < index;i ++){
+printf("%s, \n", command_args[i]);
+    }
 
     g_strfreev(command_parts);
   }
 
-  g_strv_builder_unref(cmd_builder);
   
   return command_args;
 }
@@ -209,7 +219,12 @@ static gboolean exec_thumbnailer_command(gchar **command_args) {
     NULL, G_SPAWN_DEFAULT, NULL, NULL, NULL, NULL, &wait_status, &error);
 
   if (spawned) {
-    return g_spawn_check_wait_status(wait_status, NULL);
+    //return g_spawn_check_wait_status(wait_status, NULL);
+    if ( WIFEXITED(wait_status) == 0 &&  WEXITSTATUS(wait_status) == 0 ){
+      return TRUE;
+    } else {
+      return FALSE;
+    }
   } else {
     g_warning("Error calling thumbnailer: %s", error->message);
     g_error_free(error);
```

giomatfois62:
okey fixed it, used `g_spawn_check_exit_status` to check the thumbnailer process and a simple malloc-free for the list of of arguments.  
It works for me with the new glib, using your diff I also was experiencing icons generated and not loaded

DaveDavenport:
yeah, I kinda quickly hacked it together to test while doing something else :D

giomatfois62:
I thought a bit about ways to improve performance when thumbnails are enabled.  
What I observe is the following.  
When I launch a rofi script (filebrowser or my own with custom preview) and I type the query or change page slowly, the GUI stays responsive and thumbnails are slowly (or rapidly if already in cache) loaded.  
When I quickly type or change pages, the image-fetcher jobs will pile up and make the GUI unresponsive for a bit. Once the jobs are done I can resume browsing pages or typing.  
So my idea is the following: when the user changes page or filter query, then the view is refreshed. At this point it's probably best to kill all current icon-fetcher running jobs since the user won't see those images.  
Do you think this is something reasonable that can be done with the multithread jobs implementation of rofi?

EDIT: a dirtier alternative could be to use a short duration timer on view refresh to start the icon-fetcher jobs, something like 50-100 milliseconds

DaveDavenport:
Need to think about this one.  Currently there is a threadpool and jobs are queued. It should be doable to remove jobs from the queue again if they are not yet picked up by a worker. 

If I have some time I will try to take a look at it, I am interested in seeing what is causing the slow down when the jobs pile up, maybe I can improve this.

giomatfois62:
I see there's a method in GThreadPool to query the number of unprocessed jobs `g_thread_pool_unprocessed` but I don't see something like `drop_unprocessed_jobs` that should do what we need.  
Another idea could be to queue GUI refresh jobs to the front of the queue using `g_thread_pool_move_to_front` and push icon-fetcher jobs at the back to keep the GUI responsive

DaveDavenport:
Yeah, I need to think about how to solve this.  Maybe we can set a priority and sort on them (`g_thread_pool_set_sort_function`). Then we have some control to make sure filter tasks are preferred over icon fetching tasks.

I might give this a try on next today.

DaveDavenport:
I've added (not well tested) sorting on the jobs . (hopefully right order) on the next branch.

giomatfois62:
Thanks! I tested it and my scripts are much more responsive now, even when filtering many entries quickly. There's still the issue that jobs piling up means i don't see thumbnails of the current page until all previous jobs are done, so when scrolling rapidly it can take a while, but I can live with that, until the threadpool gains the ability to cancel pending jobs (might be a good feature request to the glib folks).  

I'll show you here some of the things I'm now able to do in my scripts using the custom preview-cmd:

Program icons for a flathub applications browser
flathub

Web radio logos
radio

IPTV channels logos
tv

rss.com podcasts cover in low and high resolution
podcast

the preview script is basically just a call to wget, and i feed the urls to rofi as icons using \x00icon\x1fthumbnail://<URL>

`
#!/bin/bash

if [ -z "$1" ]; then
    exit 1
fi

nice -n 19 /usr/bin/wget -q -O "$2" "$1"
`  
I'm very happy with how it's working so far

DaveDavenport:
Very nice..  but regrettable the internal GAsyncQueue (it looks like they are using this) is not exposed. 

```c
struct _GRealThreadPool
{
  GThreadPool pool;
  GAsyncQueue *queue;
  GCond cond;
  gint max_threads;
  guint num_threads;
  gboolean running;
  gboolean immediate;
  gboolean waiting;
  GCompareDataFunc sort_func;
  gpointer sort_user_data;
};
```
We could hack around it, but we probably should not do that..  we would be better off adding a copy and modifying it for our needs.

One things I want to change is that on rofi exits, it cancels jobs in the queue instead of what it currently does (waiting till its done). 

This should be doable to add in the current implementation.

DaveDavenport:
Just two random thoughts:

* Would it be beneficial to make the download call (instead of external wget) internal?  (either via gio or libcurl or so)
* Should the normal image loader (when not loading icons from icon theme) also use and write out the thumbnails?  This might be beneficial for people making/using it as wallpaper browser and the like.

giomatfois62:
Regarding your first thought, making the download directly in rofi code would surely be a bit faster, don't know how much though, I will make some tests with both curl and gio. While introducing a new library would not be a big problem (it could even be made optional at compile time), rofi would then become closer to a little web browser, don't know if that's the direction you want the project to follow. A little side note, in my scripts I do not always have the url ready to use as icon, the preview-cmd helps allowing me for example to grep it somewhere using the entry name, or using the requested size the download an appropriate icon.  

Regarding the thumbnails generation also for normal images, that would be certainly beneficial for subsequent runs of rofi, and it allows leveraging on the work of the user file manger which most of the times already makes thumbnails for us

DaveDavenport:
You are right about the first point.. My biggest problem with rofi :smile:  I tend to think 'hmm can I do this' and then implement it before thinking if I should... (*cough* overkill theming system anybody *cough*).
I do not want rofi to be a browser, that comparison gets made way to often. 

I think the 2nd is better, esp given most of the code for it is there already with this patch.

DaveDavenport:
@lbonn  remarked (I did not know) there is an md5 implementation in glib 
https://docs.gtk.org/glib/struct.Checksum.html

I might have time Thursday to look at this.

giomatfois62:
> @lbonn remarked (I did not know) there is an md5 implementation in glib https://docs.gtk.org/glib/struct.Checksum.html
> 
> I might have time Thursday to look at this.

Thanks for pointing this out, I tested the glib GChecksum and it works good. I updated the pull request to make use of it

DaveDavenport:
That is amazingly quick

DaveDavenport:
```patch
diff --git a/meson.build b/meson.build
index bac07b0a..ff5fb2b3 100644
--- a/meson.build
+++ b/meson.build
@@ -187,7 +187,6 @@ rofi_sources = files(
         'source/timings.c',
         'source/history.c',
         'source/theme.c',
-        'source/md5.c',
         'source/rofi-icon-fetcher.c',
         'source/css-colors.c',
         'source/widgets/box.c',
@@ -220,7 +219,6 @@ rofi_sources = files(
         'include/keyb.h',
         'include/view.h',
         'include/view-internal.h',
-        'include/md5.h',
         'include/rofi-icon-fetcher.h',
         'include/helper.h',
         'include/helper-theme.h',
```

DaveDavenport:
(and enable it in recursive browser?)
```patch
diff --git a/source/modes/recursivebrowser.c b/source/modes/recursivebrowser.c
index 3e010b41..875e32cd 100644
--- a/source/modes/recursivebrowser.c
+++ b/source/modes/recursivebrowser.c
@@ -450,6 +450,10 @@ static cairo_surface_t *_get_icon(const Mode *sw, unsigned int selected_line,
   }
   if (rofi_icon_fetcher_file_is_image(dr->path)) {
     dr->icon_fetch_uid = rofi_icon_fetcher_query(dr->path, height);
+  } else if (dr->type == RFILE) {
+    gchar *_path = g_strconcat("thumbnail://", dr->path, NULL);
+    dr->icon_fetch_uid = rofi_icon_fetcher_query(_path, height);
+    g_free(_path);
   } else {
     dr->icon_fetch_uid =
         rofi_icon_fetcher_query(rb_icon_name[dr->type], height);
```

giomatfois62:
sorry, forgot about meson again xD

source/modes/recursivebrowser.c
```diff
@@ -447,12 +447,12 @@ static cairo_surface_t *_get_icon(const Mode *sw, unsigned int selected_line,
   FBFile *dr = &(pd->array[selected_line]);
   if (dr->icon_fetch_uid > 0 && dr->icon_fetch_size == height) {
     return rofi_icon_fetcher_get(dr->icon_fetch_uid);
-  }
-  if (rofi_icon_fetcher_file_is_image(dr->path)) {
```
lbonn, line 499: The `is_image` check is removed here but not in plain `filebrowser.c`.

Probably we could put more code in common between filebrowser and its recursive version (note for later, not required for this PR)

source/rofi-icon-fetcher.c
```diff
@@ -282,6 +451,64 @@ gboolean rofi_icon_fetcher_file_is_image(const char *const path) {
...
+    *thumb_size = 256;
+    thumb_path = g_strconcat(cache_dir, "/thumbnails/large/",
+        md5_hex, ".png", NULL);
+  } else if (requested_size <= 512) {
+    *thumb_size = 512;
+    thumb_path = g_strconcat(cache_dir, "/thumbnails/x-large/",
+        md5_hex, ".png", NULL);
+  } else {
+    *thumb_size = 1024;
+    thumb_path = g_strconcat(cache_dir, "/thumbnails/xx-large/",
```
lbonn, line 493: I did not have the `thumbnails/xx-large` directory on my system for some reason which blocked the generation of large previews, at least with `xapp-mp3-thumbnailer`.

We can create the target directory if it does not exist before running the thumbnailing commands.

source/rofi-icon-fetcher.c
```diff
@@ -91,6 +98,152 @@ typedef struct {
...
+      if (strcmp(command_parts[i], "%i") == 0) {
+        command_args[current_index++] = strdup(filename);
+      } else if (strcmp(command_parts[i], "%u") == 0) {
+        command_args[current_index++] = strdup(encoded_uri);
+      } else if (strcmp(command_parts[i], "%o") == 0) {
+        command_args[current_index++] = strdup(output_path);
+      } else if (strcmp(command_parts[i], "%s") == 0) {
+        char size_str[33];
+        snprintf(size_str, 33, "%d", size);
+        command_args[current_index++] = strdup(size_str);
```
lbonn: `g_strdup_printf` should be more idiomatic here.

Maybe also `g_malloc0` and `g_strdup` in the other calls for consistency?

giomatfois62:
I thought about a possible workaround to cancel unprocessed icon-fetcher jobs when changing page, we could free the tpool using `g_thread_pool_free(tpool, TRUE, FALSE)` so that it returns immediately without waiting for jobs in execution and then recreate it just after. Do you think it could be feasible?

config/config.c
```diff
@@ -45,6 +45,9 @@ Settings config = {
 
     /** Whether to load and show icons */
     .show_icons = FALSE,
+
```
lbonn: Small styling issue: there are trailing white spaces on empty lines in the whole diff. We can remove them before merging if possible.

source/modes/recursivebrowser.c
lbonn: On a related note, the existing rofi image preview code can also be extended to use standard thumbnails in the future :)

lbonn:
@giomatfois62 maybe it's a dumb question and the solution is just to install more icon packs (or I am missing some mime conf files?). Also, it's not really in scope for these changes but something to consider in future improvements.

I am not getting mime icons on (for example) mp3 files as I do not have a `audio-mpeg` icon installed. However, Adwaita has `audio-x-generic` (/usr/share/icons/Adwaita/scalable/mimetypes/audio-x-generic.svg) which would be a suitable fallback.

Reading https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#icon_lookup and https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html#names (Table 10: Standard MIME Type Icons), should rofi look for these generic icons in case the first lookup fails? I don't know how other apps like Thunar are handling this.

giomatfois62:
> @giomatfois62 maybe it's a dumb question and the solution is just to install more icon packs (or I am missing some mime conf files?).

It's not dumb, I'm also missing some icons because the mimetype is unresolved or too generic (in my case, .iso images, that are displayed correctly in dolphin and thunar). It depends both on the icon theme in use and in the inner workings of the library used by rofi to fetch these icons by name (the function is `nk_xdg_theme_get_icon`).  
File managers use some many workarounds to always display something, in this pull request I tried to keep the workflow as generic and standard as possible.

source/modes/recursivebrowser.c
DaveDavenport: > Probably we could put more code in common between filebrowser and its recursive version (note for later, not required for this PR)

Agree, the recursive browser was something I did to see if I  could again (a file browser that was very very quick for users whole home directory and non blocking of the UI). 
Lets put this on the todo list.

DaveDavenport:
> > @giomatfois62 maybe it's a dumb question and the solution is just to install more icon packs (or I am missing some mime conf files?).
> 
> It's not dumb, I'm also missing some icons because the mimetype is unresolved or too generic (in my case, .iso images, that are displayed correctly in dolphin and thunar). It depends both on the icon theme in use and in the inner workings of the library used by rofi to fetch these icons by name (the function is `nk_xdg_theme_get_icon`). File managers use some many workarounds to always display something, in this pull request I tried to keep the workflow as generic and standard as possible.

@sardemff7 ^^

config/config.c
DaveDavenport: I recommend running clang-format with default settings, that should in general be the style used.

giomatfois62:
I'm trying to stop icon fetcher jobs when the current page is changed, either by scrolling entries or pressing page up/down. This way I don't have to wait for pending jobs when i change page quickly, and the thumbnails for the current page are immediately generated.  
To do this  I simply delete and recreate the icon fetching threadpool on every page change (in listview.c).  
Some jobs will then be discarded because they are still pending in the queue.  
Discarded jobs will appear in the icon_cache hashmaps with their name and uid, but won't have a valid image surface generated (as is expected).
The problem I'm facing is, when I go back to a previous page, I don't see some thumbnails until I hover their entries.  
What do you think is causing this? I'm looking at the "query_done" attribute of the IconFetcherEntry to check if a thumbnail was generated succesfully or discarded when deleting the threadpool, perhaps there's a better way?  
Below my current diff

```diff
diff --git a/source/rofi-icon-fetcher.c b/source/rofi-icon-fetcher.c
index c0019148..80ac2fec 100644
--- a/source/rofi-icon-fetcher.c
+++ b/source/rofi-icon-fetcher.c
@@ -737,6 +737,8 @@ uint32_t rofi_icon_fetcher_query_advanced(const char *name, const int wsize,
        iter = g_list_next(iter)) {
     sentry = iter->data;
     if (sentry->wsize == wsize && sentry->hsize == hsize) {
+      if (!sentry->query_done)
+        g_thread_pool_push(tpool, sentry, NULL);
       return sentry->uid;
     }
   }
@@ -775,6 +777,8 @@ uint32_t rofi_icon_fetcher_query(const char *name, const int size) {
        iter = g_list_next(iter)) {
     sentry = iter->data;
     if (sentry->wsize == size && sentry->hsize == size) {
+      if (!sentry->query_done)
+        g_thread_pool_push(tpool, sentry, NULL);
       return sentry->uid;
     }
   }
diff --git a/source/view.c b/source/view.c
index 29ca2307..2c84994f 100644
--- a/source/view.c
+++ b/source/view.c
@@ -2664,7 +2664,7 @@ void rofi_view_workers_initialize(void) {
 }
 void rofi_view_workers_finalize(void) {
   if (tpool) {
-    g_thread_pool_free(tpool, TRUE, TRUE);
+    g_thread_pool_free(tpool, TRUE, FALSE);
     tpool = NULL;
   }
 }
diff --git a/source/widgets/listview.c b/source/widgets/listview.c
index 2e333cb6..5e1ca43c 100644
--- a/source/widgets/listview.c
+++ b/source/widgets/listview.c
@@ -284,6 +284,11 @@ static unsigned int scroll_per_page(listview *lv) {
         (lv->max_elements > 0) ? (lv->selected / lv->max_elements) : 0;
     offset = page * lv->max_elements;
     if (page != lv->cur_page) {
+
+      g_warning("Page change, resetting icon fetcher threadpool");
+      rofi_view_workers_finalize();
+      rofi_view_workers_initialize();
+
       lv->cur_page = page;
       lv->rchanged = TRUE;
     }

```

DaveDavenport:
I'll take a look at it later this week.

giomatfois62:
> I'll take a look at it later this week.

Thanks Dave, I'll keep messing up in the meantime

giomatfois62:
Found the issue! When rofi updates an item and request its icon, it checks if it has an icon_fetch_uid greater than zero and returns the surface directly (The same is done in filebrowser and recursivebrowser):
https://github.com/davatorium/rofi/blob/1c8159c4c893ff558b28345a82270a142da3fc5d/source/modes/dmenu.c#L718C3-L720C4  
Now that I destroy and rebuild the icon fetcher threadpool, entries can have a valid icon_fetch_uid without the corresponding surface because the query was not completed.  
What I did to workaround the problem was the following: 
- added a boolean field "query_started" to IconFetcherEntry, this is set to TRUE once the worker thread starts its job, then its guaranteed that an icon will eventually be produced, otherwise the job could get discarded before starting.
- commented the lines pasted above, this way I force a lookup in the cache map to retrieve the uid and also check if the query was started. If it was not started, then it means that the job was discarded when deleting the threadpool and i submit it again.  

Now it works wonderfully, I'll clean up the code and commit so you can take a look for yourself

DaveDavenport:
Great!.

I've been looking at what we can do about the threadpool.. I am not sure if destroying/recreating it is the best (This might also cause a memory leak as some structs that go in are allocated and some are stacked. I will fix this (easy enough), but did not have time yet).

Ideally I would like to be able to clear out the queue instead of destroy /recreate..   everything we need exists, except that glib hides that field .

giomatfois62:
ok thanks, let me know if I can help with something

sardemff7:
> I am not getting mime icons on (for example) mp3 files as I do not have a audio-mpeg icon installed. However, Adwaita has audio-x-generic (/usr/share/icons/Adwaita/scalable/mimetypes/audio-x-generic.svg) which would be a suitable fallback.
> 
> Reading https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#icon_lookup and https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html#names (Table 10: Standard MIME Type Icons), should rofi look for these generic icons in case the first lookup fails? I don't know how other apps like Thunar are handling this.

\+

> It's not dumb, I'm also missing some icons because the mimetype is unresolved or too generic (in my case, .iso images, that are displayed correctly in dolphin and thunar). It depends both on the icon theme in use and in the inner workings of the library used by rofi to fetch these icons by name (the function is nk_xdg_theme_get_icon).
File managers use some many workarounds to always display something, in this pull request I tried to keep the workflow as generic and standard as possible.

Reading the naming spec, and refreshing my memory on the lookup algorithm, I see an easy enough solution

I’ll add an option on the lookup function to support names fallbacks, just like the sound theme spec (lucky me, I made the algorithm purely generic so it’s just a matter of giving an array of many names instead of an array of one)
The sad part is that I’ll have to special-case the MIME stuff because someone decided that instead of stripping the post-`-` part of the name we’d need `-x-generic` which is dumb honestly but whatever

So at some point in the future these will just work™ in rofi (with a submodule update + adding a flag) and no need to add workarounds here (yet, there may be a case that doesn’t make sense in libnkutils though we won’t know until we stumble on it)

DaveDavenport:
I've been planning to bump glib  up to ubuntu 22.04 version (2.72). 
Because of deprecation we also need to bump meson, so it might be good to just make a big step-change.

DaveDavenport:
on ~~main~~ next branch I've added free'ing on the threadpool.. so destroying it while items in queue should not leak anymore.

Not well tested, I've been busy.

giomatfois62:
Thanks, I'll test a bit this weekend

giomatfois62:
I'm getting segfaults when switching pages, so I looked a bit deeper at your changes. A couple of things I noticed: 
- in `rofi_thread_pool_state_free` I see you cast the pointer to thread_state, but shouldn't it be the IconFetcherEntry? That's the struct passed to the threadpool when calling g_thread_pool_push
- the `rofi_icon_fetcher_entry_free` that you pass as a free_func for the threadpool is already called by the `rofi_icon_fetcher_data->icon_cache` cache hashmap on its entries. This hashmap is populated before passing requests to the threadpool, so there shouldn't be any leaks when destroying the threadpool, any unfinished job has its IconFetcherEntry stored in cache and it gets destroyed when rofi quits. I agree it's a bit wasteful to leave them there, but without an actual icon loaded the memory used by an entry is basically only its name string, size and some booleans. 

I'll keep testing, there could be other things wrong in my code triggering these crashes

DaveDavenport:
>in rofi_thread_pool_state_free I see you cast the pointer to thread_state, but shouldn't it be the IconFetcherEntry? That's the struct passed to the threadpool when calling g_thread_pool_push

The first element of IconFetcherEntry should be thread_state,  therefor it is a valid cast.
There are multiple data types pushed to the thread-pool so we need to have this abstraction.

You are right on the 2nd item..  in my quick rush I completely overlooked that.
So in the 'free' function we pass to the `thread_state` should we remove the entry from the icon_cache ? or indicate in some way that on a next run it should be re-requested?

DaveDavenport:
pushed a fix and a placeholder.

hmm why can't it auto merge that.. odd.

giomatfois62:
> indicate in some way that on a next run it should be re-requested? 

Something like this is done in my pull requested by adding a new member "query_started" to IconFetcherEntry. If this is FALSE, then the icon fetch job is submitted again to the threadpool

DaveDavenport:
> > indicate in some way that on a next run it should be re-requested?
> 
> Something like this is done in my pull requested by adding a new member "query_started" to IconFetcherEntry. If this is FALSE, then the icon fetch job is submitted again to the threadpool

aah great.

DaveDavenport:
What happens if the same query is executed twice? before the first is handled.  is it re-added to the queue? 

What about (damn indenting changed):

```patch
diff --git a/source/rofi-icon-fetcher.c b/source/rofi-icon-fetcher.c
index a6ddbcb0..99620ce0 100644
--- a/source/rofi-icon-fetcher.c
+++ b/source/rofi-icon-fetcher.c
@@ -246,8 +246,9 @@ static gboolean rofi_icon_fetcher_create_thumbnail(const gchar *mime_type,
 }

 static void rofi_icon_fetch_thread_pool_entry_remove(gpointer data) {
-  IconFetcherNameEntry *entry = (IconFetcherNameEntry *)data;
+  IconFetcherEntry *entry = (IconFetcherEntry *)data;
   // Mark it in a way it should be re-fetched on next query?
+  entry->query_started = FALSE;
 }

 static void rofi_icon_fetch_entry_free(gpointer data) {
@@ -535,8 +536,6 @@ static void rofi_icon_fetcher_worker(thread_state *sdata,
   const gchar *icon_path;
   gchar *icon_path_ = NULL;

-  sentry->query_started = TRUE;
-
   if (g_str_has_prefix(sentry->entry->name, "thumbnail://")) {
     // remove uri thumbnail prefix from entry name
     gchar *entry_name = &sentry->entry->name[12];
@@ -761,7 +760,7 @@ uint32_t rofi_icon_fetcher_query_advanced(const char *name, const int wsize,
   sentry->hsize = hsize;
   sentry->entry = entry;
   sentry->query_done = FALSE;
-  sentry->query_started = FALSE;
+  sentry->query_started = TRUE;
   sentry->surface = NULL;

   entry->sizes = g_list_prepend(entry->sizes, sentry);
@@ -804,7 +803,7 @@ uint32_t rofi_icon_fetcher_query(const char *name, const int size) {
   sentry->hsize = size;
   sentry->entry = entry;
   sentry->query_done = FALSE;
-  sentry->query_started = FALSE;
+  sentry->query_started = TRUE;
   sentry->surface = NULL;

   entry->sizes = g_list_prepend(entry->sizes, sentry);
```

DaveDavenport:
hmm there is a weird thing with thread-pool..   the free method gets a invalid pointer (0x1) passed from time to time..  I cannot see why (yet).

giomatfois62:
> hmm there is a weird thing with thread-pool.. the free method gets a invalid pointer (0x1) passed from time to time.. I cannot see why (yet).

That's what is giving segfaults, the only workaround for me was to comment it all.  

> What about (damn indenting changed):  

This is good, it avoids possible race conditions since I was flipping the boolean in a different thread. Now it just needs the threadpool free_func to properly behave :)

DaveDavenport:
I currently do not see why.
* structures pushed in for filtering are waited on blocking the main thread. So the moment the pool is destroyed I know for sure that these are not in the queue.
* Structures pushed into the queue from the icon-fetcher are not free'ed. (only free is by icon_cache being cleared or if we push in a new token with the same name, but we guard for this.)

In any case I would not expect to see a pointer with value '1'.

DaveDavenport:
`g_thread_pool_free(tpool, TRUE, TRUE);` seems to fix the issue.

With wait set to false, if I gdb and step back,  g_async_queue_unref inside g-thread-pool gets a NULL pointer... odd.

The free function is not called from the same thread as `g_thread_pool_free`..  now that is odd.

DaveDavenport:
```c
static void
g_thread_pool_free_internal (GRealThreadPool* pool)
{
  g_return_if_fail (pool);
  g_return_if_fail (pool->running == FALSE);
  g_return_if_fail (pool->num_threads == 0);

  /* Ensure the dummy item pushed on by g_thread_pool_wakeup_and_stop_all() is
   * removed, before it’s potentially passed to the user-provided
   * @item_free_func. */
  g_async_queue_remove (pool->queue, GUINT_TO_POINTER (1));

  g_async_queue_unref (pool->queue);
  g_cond_clear (&pool->cond);

  g_free (pool);
}
```

well..  ok.. But I think I am catching the 1 passed here.

DaveDavenport:
Ok. the above fix was added in 2.79.. and I am running older

https://github.com/GNOME/glib/commit/b402f66c07a8a248c0d8f6f9f82279939190fbff

DaveDavenport:
```patch
diff --git a/source/rofi-icon-fetcher.c b/source/rofi-icon-fetcher.c
index a6ddbcb0..99620ce0 100644
--- a/source/rofi-icon-fetcher.c
+++ b/source/rofi-icon-fetcher.c
@@ -246,8 +246,9 @@ static gboolean rofi_icon_fetcher_create_thumbnail(const gchar *mime_type,
 }
 
 static void rofi_icon_fetch_thread_pool_entry_remove(gpointer data) {
-  IconFetcherNameEntry *entry = (IconFetcherNameEntry *)data;
+  IconFetcherEntry *entry = (IconFetcherEntry *)data;
   // Mark it in a way it should be re-fetched on next query?
+  entry->query_started = FALSE;
 }
 
 static void rofi_icon_fetch_entry_free(gpointer data) {
@@ -535,8 +536,6 @@ static void rofi_icon_fetcher_worker(thread_state *sdata,
   const gchar *icon_path;
   gchar *icon_path_ = NULL;
 
-  sentry->query_started = TRUE;
-
   if (g_str_has_prefix(sentry->entry->name, "thumbnail://")) {
     // remove uri thumbnail prefix from entry name
     gchar *entry_name = &sentry->entry->name[12];
@@ -761,7 +760,7 @@ uint32_t rofi_icon_fetcher_query_advanced(const char *name, const int wsize,
   sentry->hsize = hsize;
   sentry->entry = entry;
   sentry->query_done = FALSE;
-  sentry->query_started = FALSE;
+  sentry->query_started = TRUE;
   sentry->surface = NULL;
 
   entry->sizes = g_list_prepend(entry->sizes, sentry);
@@ -804,7 +803,7 @@ uint32_t rofi_icon_fetcher_query(const char *name, const int size) {
   sentry->hsize = size;
   sentry->entry = entry;
   sentry->query_done = FALSE;
-  sentry->query_started = FALSE;
+  sentry->query_started = TRUE;
   sentry->surface = NULL;
 
   entry->sizes = g_list_prepend(entry->sizes, sentry);
index 15c72865..4965c306 100644
--- a/source/view.c
+++ b/source/view.c
@@ -2652,7 +2652,7 @@ static int rofi_thread_workers_sort(gconstpointer a, gconstpointer b,
 }
 
 static void rofi_thread_pool_state_free(gpointer data) {
-  if (data) {
+  if (data && GPOINTER_TO_UINT(data) != 1) {
     thread_state *ts = (thread_state *)data;
     if (ts->free) {
       ts->free(data);
```

This seems to fix it for me. 

There is one change I would like to make, I do not like that listview calls  the destroy/create method from view.c on the thread pool.
This breaks the abstraction that is there..  

Maybe for now lets make a callback on the listview for a page change and then call the destroy/create inside view.c.

giomatfois62:
XD the rogue dummy item!! Great work Dave, I'll update the PR as soon as I get some free time

DaveDavenport:
pushed fix to `next` branch.

giomatfois62:
I added the callback on the listview for page changes, all working well in my tests

DaveDavenport:
I guess it is time to merge this. @lbonn should we merge this for next release or lift over the release and try to do a quick follow up release.

lbonn:
@DaveDavenport hm good question... In the notes, the filebrowser is presented as experimental so probably it is fine to include it in the release as such.

It would just need some light documentation I suppose.

giomatfois62:
I can add documentation too, where should I put it? My guess was in rofi-theme

DaveDavenport:
I think this is a good location: https://github.com/davatorium/rofi/blob/next/doc/rofi-theme.5.markdown#icon-handling
To add some information. If it needs a large explanation (to explain setup, etc), we might want to have a separate manpage.


I am still wondering if we should make this the 'default' behaviour for loading images, so we re-use preview other programs make.. The thing that worries me a little is that I had to tweak app-armour on my pcs to make it work.

giomatfois62:
I can't help much with that at the moment, I use Slackware and it doesnt come with neither apparmor nor selinux. I can test a bit at work with fedora/Ubuntu in the coming days if you can wait a bit for the new release

DaveDavenport:
No rush, I am crazy busy too.
I will test it too on ubuntu/debian and can update the document.

giomatfois62:
I'm investigating the apparmor permission issues. Looking at the gnome code [here](https://gitlab.gnome.org/GNOME/gnome-desktop/-/blob/master/libgnome-desktop/gnome-desktop-thumbnail-script.c) apparently the thumbnailers programs (the one specified in the Exec key of .thumbnailer files) are not trusted by apparmor/selinux/flatpak etc. and are restricted to write inside a sandbox, that is a temporary folder created in /tmp/. The generated images are then loaded back in memory and saved in the standard thumbnails paths ($HOME/.cache/thumbnails/...).  
Now I want to study how xfce does it (with their tumblerd implementation), kde does everything by itself using dolphin plugins and doesnt look for .thumbnailer files in the system like we do

giomatfois62:
can you try adding the following apparmor profile to /etc/apparmor.d/usr.bin.rofi  

```
#vim:syntax=apparmor
# AppArmor policy for rofi

#include <tunables/global>

/usr/bin/rofi {
    #include <abstractions/base>

    # TCP/UDP network access for NFS
    network inet  stream,
    network inet6 stream,
    network inet  dgram,
    network inet6 dgram,

    /usr/bin/rofi mr,

    @{HOME}/ r,
    @{HOME}/** rw,
    owner @{HOME}/.cache/thumbnails/** rw,
}
```

then run `apparmor_parser  -r /etc/apparmor.d/usr.bin.rofi` to reload the rule. This assumes that rofi binary is in /usr/bin.
I adapted the profile of evince-thumbnailer together with a snippet taken from [here](https://nevesnunes.github.io/blog/2022/11/12/No-thumbnails-for-you.html)

DaveDavenport:
I will test this somewhere this week.. Thanks a lot

DaveDavenport:
did not forget, will get back to testing.

DaveDavenport:
I think one of the important things left todo is test the above, but mostly document this feature. 
Anybody willing to take this task?

giomatfois62:
I can write some documentation, perhaps also pointing to the apparmor rule if it works for everyone (I only tested it on my work pc), where do you think is best to put it?

DaveDavenport:
Currently we have a small section here on icon loading: https://github.com/davatorium/rofi/blob/next/doc/rofi-theme.5.markdown#icon-handling

However it might be good to create a new manpage for this, given the questions we get it could not hurt to expand the current section  and add the new info..

@lbonn  what do you think?

lbonn:
I agree, it is probably worthy of its own man page, with a link (and adjustments) in the existing section.

`rofi-filebrowser.5` or `rofi-thumbnails.5`? If we pick filebrowser, we also have a space to document the mode(s) with a little bit more details.

DaveDavenport:
I would go for rofi-thumbnails, because I think it would be nice to (in a next version) use this more generic .

giomatfois62:
I finally got some time to write documentation of the new feature, let me know what you think about it

lbonn:
@giomatfois62 I've added a few tweaks on the docs.

The man page does not display too nicely with the huge `MimeType=` line but that's probably fine :)

giomatfois62:
Thank you very much!

DaveDavenport:
Is it ready to be merged, or are there more updates you want to make before I hit the button?

giomatfois62:
No more updates from me, ready to merge if you want

DaveDavenport:
Thanks!!!! merged

giomatfois62:
Thankyou!

PR #1955, opened 2024-03-02
[Doc] Switch to pandoc and remove generated files

lbonn:


lbonn:
Oops this whole lua transformation needs pandoc 2.17 and 22.04 is stuck on 2.9...

This method has the advantage of giving a good layout on manpages, github preview and mkdocs. But the lua filter is a bit complicated.

I also tried to rewrite the man files with definition lists but they don't render very well. Mkdocs is supposed to handle them but in practice it breaks if a code block is nested inside a definition.

Solutions?

* wait for github to have 24.04 as ubuntu-latest...
* `####` for command definitions, not as pretty in makefiles
* graceful degradation if pandoc < 2.17

EDIT: went with the graceful degradation!

lbonn:
```
 ERROR: files left in build directory after distclean:
./=
```

hmm I guess something is wrong with my autoconf. The file appears when running `./configure`


EDIT: yes it was the wrong version check for pandoc. Gave up on it...

PR #1969, opened 2024-03-25
Add border_width to window coordinates

ortango:
In reference to #1968, this patch is to account to border_width for window coordinates when displaying on the currently focused window (`-monitor -2`).

Rather then add border width to the window's height and width, we are moving the x, y offsets. _100% of the window would be everything but the border_. I looked through xlib manual and icccm, but did not come across a defacto standard for this.

### DIFF
--- a/include/xcb.h
+++ b/include/xcb.h
@@ -209,8 +209,6 @@ typedef enum {
   WM_DO_NOT_CHANGE_CURRENT_DESKTOP = 1,
   /** PANGO WORKSPACE NAMES */
   WM_PANGO_WORKSPACE_NAMES = 2,
-  /** Root window offset (for bspwm) */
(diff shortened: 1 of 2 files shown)


ortango:
This changes current behavior for both reparenting and non-reparenting wm's. In both cases 100% x/y offsets and height/width of the window fills the area inside the border/decorations.

_____

so these changes are a little to much. i tested the following wm's : bspwm, dwm, i3 and everything is fine with -m -{2,4}.
~~testing xmonad (after seeing the xmonad issue/commit) - both 2 and 4 are broken.~~ did not realize that xmonad's ewmh support was optional. after enabling it this is working.

so, this patch also fixes -m -2 for other nonreparenting wm's aside from bspwm.

DaveDavenport:
Thanks!

PR #1988, opened 2024-06-09
Add -transient-window mode

pijulius:
>
> - [ X] This PR targets the `next` branch and not `master`
> - [ -] If your PR is a work in progress, include [WIP] in its title
> - [ X] Its commits' summaries are reasonably descriptive
> - [ X] You've described what this PR addresses below
> - [ -] You've included links to relevant issues, if any with `#issue_num`
> - [ -] You've deleted this template
>

This will add a new mode for the rofi window so when you run it with -transient-window

instead of
-normal-window

it will get the currently focused window and then use that as the parent window and set itself as transient of that window.  It is useful so that when you  use a keyboard shortcut to start rofi it will always popup on the window you are working on and so it doesn't make you refocus/move your head on large monitors :)

### DIFF
--- a/doc/rofi.1.markdown
+++ b/doc/rofi.1.markdown
@@ -302,6 +302,12 @@ Use Pango markup to format output wherever possible.
 Make **rofi** react like a normal application window. Useful for scripts like
 Clerk that are basically an application.
 
+`-transient-window`
+
+Make **rofi** react like a modal dialog that is transient to the currently
+focused window. Useful when you use a keyboard shortcut to run and show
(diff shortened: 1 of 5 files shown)


DaveDavenport:
Good idea!

Can you include it in the manpages? then can will merge it.

pijulius:
Thanks @DaveDavenport ! happy to see you like it too, not sure about the docs, this is what you had in mind?:
https://github.com/davatorium/rofi/commit/f621786680c9e65efb714465ee532162889a522d

PR #1999, opened 2024-07-09
Fix ime and enable build option

amosbird:
This is a follow-up fix PR to https://github.com/davatorium/rofi/pull/1735 which makes fcitx5 work perfectly in rofi.


amosbird:
@DaveDavenport Oops. It somehow doesn't work with modifiers. That means, when invoking rofi using mod + keys, it fails to release the mod key and disrupts interaction. I'll submit another PR to avoid forwarding modifiers ftm.

PR #2000, opened 2024-07-09
Better XIM interaction by filtering modifiers

amosbird:
After diagnosing both the server and client sides of xcb-imdkit, specifically rofi and fcitx5, it has been determined that modifiers are causing the slowdown. This is a follow-up to PR #1999.

A new dependency, `xcb-keysyms`, has been added to properly recognize modifier keys.

I only performed some basic tests based on my daily workflow, and it worked smoothly. Let's see if this fix can stand the test of time.

### DIFF
--- a/include/xcb-internal.h
+++ b/include/xcb-internal.h
@@ -36,6 +36,7 @@
 #include <libgwater-xcb.h>
 #include <xcb/xcb.h>
 #include <xcb/xcb_ewmh.h>
+#include <xcb/xcb_keysyms.h>
 
 #include <nkutils-bindings.h>
 
@@ -49,6 +50,7 @@ struct _xcb_stuff {
 #ifdef XCB_IMDKIT
   xcb_xic_t ic;
   xcb_xim_t *im;
+  xcb_key_symbols_t *syms;
 #endif
   xcb_ewmh_connection_t ewmh;
   xcb_screen_t *screen;
--- a/meson.build
+++ b/meson.build
@@ -79,6 +79,7 @@ deps += [
     dependency('xcb-randr'),
     dependency('xcb-cursor'),
     dependency('xcb-xinerama'),
+    dependency('xcb-keysyms'),
     dependency('cairo-xcb'),
     dependency('libstartup-notification-1.0'),
 ]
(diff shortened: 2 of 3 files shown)


xsy420:
I have tried this pull and [#1999](https://github.com/davatorium/rofi/pull/1999)
before these two commits, fcitx5 can show and change input method by using
```shell
meson setup --reconfigure -Dimdkit=true --prefix=~/.local/ build
ninja -C build
```
but since [#1999](https://github.com/davatorium/rofi/pull/1999), these two commands can't build a rofi that support fcitx5 showing
I have also tried building command in [archlinux-rofi](https://gitlab.archlinux.org/archlinux/packaging/packages/rofi/-/blob/main/PKGBUILD?ref_type=heads)
I have `xcb-imdkit` and `xcb-keysyms` installed
what am I missing? and what's your building command to build a rofi that makes fcitx5 work perfectly?

DaveDavenport:
Do you have any errors it reports?  'it does not build' gives me very little to go on.

For me latest git (that includes these patches) builds fine.

xsy420:
building process goes fine but after I run rofi -show drun and press my fcitx5 input method changing key-binding fcitx5 doesn't show. before these two commits, it can.
[meson.log](https://github.com/user-attachments/files/16432101/meson.log)
[ninja.log](https://github.com/user-attachments/files/16432102/ninja.log)

DaveDavenport:
aah, I thought it did not build with imdkit enabled. I  know to little about this to help, @amosbird do you know?

amosbird:
@xsy420 What key binding do you use to toggle the input method?

xsy420:
when I downgrade xcb-imdkit to 1.0.2
[meson.log](https://github.com/user-attachments/files/16432322/meson.log)
left shift key to toggle

xsy420:
so it is key-binding conflict?

amosbird:
Modifier-alone key bindings won't work after this PR. However, there should be a way to improve this by allowing such events to pass through. I will look into this further.

xsy420:
maybe try to read [HotKey/TriggerKeys] value in $XDG_CONFIG_DIR/fcitx5/config?

PR #2015, opened 2024-08-02
Follow up #2000

amosbird:
Address https://github.com/davatorium/rofi/pull/2000#issuecomment-2259575998

It seems we can work around the limitation of XCB XIM by introducing a short sleep before forwarding the release event of modifier keys. This change should enable all modifier keys to function correctly with the input method.

@DaveDavenport, could you please test this solution again? I believe it will also resolve the ctrl-u issue you experienced earlier.
@xsy420, could you please check if this fix works for you as well?

### DIFF
--- a/source/xcb.c
+++ b/source/xcb.c
@@ -1375,14 +1375,9 @@ static void main_loop_x11_event_handler_view(xcb_generic_event_t *event) {
     xcb_key_press_event_t *xkpe = (xcb_key_press_event_t *)event;
 #ifdef XCB_IMDKIT
     if (xcb->ic) {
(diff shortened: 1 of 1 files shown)


DaveDavenport:
both switching with left shift and ctrl-u seems to work for me. Thanks!

(odd/shame we need a 5ms timeout.. )

amosbird:
> both switching with left shift and ctrl-u seems to work for me. Thanks!
> 
> (odd/shame we need a 5ms timeout.. )

Yeah. I suppose it's related to some packet coalescing issue. Fortunately, we only need to delay for the release of modifiers, which should have minimal impact.

xsy420:
yeah, it works for me

Issue #2081, opened 2025-01-16
[BUG] Buffer overflow in rofi -e -


faulesocke:
### Rofi version (rofi -v)

1.7.7

### Configuration

na

### Theme

na

### Launch command

`rofi -e -`

### Step to reproduce

1. Launch `rofi -e -`
2. Pipe some text to stdin. Works best with > 30 chars.
3. Observe output on screen

### Expected behavior

Only the piped-in text should be shown

### Actual behavior

Sometimes extra garbage is added to the end of the text

### Additional information

This happens in https://github.com/davatorium/rofi/blob/next/source/rofi.c#L820

After the while loop, there should be a
```c
msg[index] = 0;
```

Sidenote: The code is also suboptimal, since even when only one read() ever returns data, it reallocates the buffer at least once.

But the code is simple and this does not really matter.

### Using wayland display server protocol

- [X] No, I don't use the wayland display server protocol

### I've checked if the issue exists in the latest stable release

- [X] Yes, I have checked the problem exists in the latest stable version

faulesocke:
For reproducing this in the shell, do something like:
```bash
echo -n "foobar bluuuu blaaa blubb blaaah" | rofi -e -
```

DaveDavenport:
Merged fix, thanks.

PR #2082, opened 2025-01-16
Fix buffer overflow in rofi -e after reading from stdin

faulesocke:
See #2081 for more info. This fixes the problem. Tested it.

It is always possible to write at `msg[index]` since there will be exactly 1024 bytes free after the read data, which ends at `index -1`.


DaveDavenport:
Thanks!

PR #2087, opened 2025-01-21
Add option to enable the `kb-delete-entry` in script mode

douo:
What this PR does:

- Enable the `kb-delete-entry` keybinding in scripts through a new `use-delete` option.
  - Default is `false` to maintain backward compatibility and avoid breaking existing behavior.
- Updated documentation to reflect the new state for entry deletion.
- Modified the script environment example to include the new option.

Reason:

This change enhances user control over entry management in scripts. Specifically, for clipboard management, providing the ability to delete clipboard entries is an important feature.

### DIFF
--- a/Examples/test_script_env.sh
+++ b/Examples/test_script_env.sh
@@ -9,7 +9,6 @@ fi
 echo -en "\x00no-custom\x1ffalse\n"
 echo -en "\x00data\x1fmonkey do, monkey did\n"
 echo -en "\x00use-hot-keys\x1ftrue\n"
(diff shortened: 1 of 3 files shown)


DaveDavenport:
I don't think we need the 'use-delete' flag.  I don't think it would collide/break any functional to have it always 'on'.

The normal hotkeys do (as they are used to quick-switch between modes), but this should not be an issue.

douo:
Thank you for the feedback! I’ve updated the PR to remove the `use-delete` flag as  your suggestion. The `kb-delete-entry` keybinding is now always enabled in script mode.

PR #2099, opened 2025-02-19
[WIP] Just to look at diff.

DaveDavenport:
>
> - [ ] This PR targets the `next` branch and not `master`
> - [ ] Its commits' summaries are reasonably descriptive
> - [ ] You've described what this PR addresses below
> - [ ] You've included links to relevant issues, if any with `#issue_num`
>

Your description here...


lbonn:
@DaveDavenport

Big outline:

What I did not want was to just add #ifdef of ifs in every function :smile: 

* xcb.c is renamed to display.c and the two implementations are in `xcb/display.c` and `wayland/display.c`. display-internal.h describes the interface they implement and plain display.c just proxies between the two.
* the split of view.c is not as clean: generic parts are kept in view.c but everything that was xcb-specific or differed too much from wayland was moved in `xcb/view.c` and `wayland/view.c`. The functions which are proxied this way are at the bottom of view.c
* there are actually a small number of ifdefs in view.c for some simple cases

For view, ideally wayland/view.c and xcb/view.c should disappear and some of their functionality can be moved to the corresponding display.c and/or another object.
I haven't reviewed exactly what could be put in common recently but I suspect there are at least some low hanging fruits.

Though they sizeable, around 1000 lines for xcb and half for wayland and both display.c are ~2000 lines. So, moving everything there would probably make a mess.

Issue #2102, opened 2025-02-19
[Request] Option to Exclude Categories from drun.


DaveDavenport:
### Discussed in https://github.com/davatorium/rofi/discussions/2101

<div type='discussions-op-text'>

<sup>Originally posted by **J0hannes101** February 19, 2025</sup>
Currently, ```-drun-categories``` allows filtering by inclusion, but as far as I know, there is no option to exclude entries based on their category.

It would be useful to have a flag like ```-drun-exclude-categories "Game"``` to hide unwanted categories without modifying .desktop files or using scripts.</div>

J0hannes101:
I think we should close this issue since #2104 was merged.

DaveDavenport:
Yes, thanks for patch.

PR #2104, opened 2025-02-20
Add option to exclude categories from drun #2102

J0hannes101:
This PR adds a ```-drun-exclude-categories``` option to exclude entries from drun based on their category.


source/modes/drun.c
```diff
@@ -722,6 +723,17 @@ static void read_desktop_file(DRunModePrivateData *pd, const char *root,
     }
   }
 
+  if (pd->exclude_categories) {
+    categories = g_key_file_get_locale_string_list(
```
DaveDavenport: If both show_categories and exclude_categories is used, we have a memory leak here. 
We should probably add a check here around it to only get Categories if categories is NULL.

DaveDavenport:
Looks good. Can you also update the manpage with this new option?

J0hannes101:
Anything else to do?

DaveDavenport:
thanks.

PR #2106, opened 2025-02-24
Add theming to hide text cursor if no text is put in

J0hannes101:
This PR adds a theming option to hide the text cursor when the input is empty, as discussed in #2105.

It works as expected, but we still need to decide on:

- The naming of the theming option (currently ``hide-empty-cursor: true``)
- Which widget should be used to place the ``hide-empty-cursor`` option in a theme (currently in ``entry{}``)

Also need to update the docs.

### DIFF
--- a/doc/rofi-theme.5.markdown
+++ b/doc/rofi-theme.5.markdown
@@ -132,18 +132,18 @@ element-text {
 └─────────────────────────────────────────────────────────────────────┘ 
 ```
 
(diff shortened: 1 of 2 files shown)


DaveDavenport:
The name of the widget is the right place to place this option on. 
Need to think about naming.. it now sounds like hiding an empty cursor, not hiding cursor on empty textbox.
Maybe 'hide-cursor-on-empty'?

DaveDavenport:
Thanks.

PR #2108, opened 2025-03-01
[scrollbar] Add theming for rounded corners

J0hannes101:
This PR adds a new ``rounded-corners`` option in the ``scrollbar{ }`` category for theming.

Example: 
```
scrollbar {
    padding: 0px 0px 0px 10px;  
    rounded-corners: true;
    handle-width: 6px ;
    handle-color: red;
}
```

### DIFF
--- a/doc/rofi-theme.5.markdown
+++ b/doc/rofi-theme.5.markdown
@@ -1011,6 +1011,7 @@ The following properties are currently supported:
 - **handle-width**:        distance
 - **handle-color**:        color
 - **border-color**:        color
+- **rounded-corners**:     boolean for rounded scrollbar
 
 ### box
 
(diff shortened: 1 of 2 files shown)


DaveDavenport:
Thanks for the patch.
I had on my todo list, to turn the handle as a subwidget, so it would inherit all theme theming options of a normal 'widget' (including the rounded corners). 

Maybe this solution is easier and sufficient. 
Would ' rounded-corners' better be 'handle-rounded-corners'? so it matches with the other properties?

J0hannes101:
So i should rename it ``handle-rounded-corners`` for now?

DaveDavenport:
It would make it more clear its on the handle, and not the widget itself (that is 'border-radius'  where you can specify it for each corner)

J0hannes101:
@DaveDavenport is everything  good for merging, or did i miss something?

DaveDavenport:
No, just have been very busy.

PR #2115, opened 2025-03-12
added a transparent theme

killertofus:
this is a transparent theme

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -408,6 +408,7 @@ install_data(
     'themes/iggy.rasi',
     'themes/material.rasi',
     'themes/iggy.jpg',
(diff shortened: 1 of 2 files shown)


DaveDavenport:
It needs a copyright notice/user at the top (see other themes). Otherwise debian will not package it.

DaveDavenport:
Theme did not work for me when loaded as theme. Tried to fix this by loading default theme first, before overriding color scheme.

killertofus:
@DaveDavenport

DaveDavenport:
@killertofus 
image

killertofus:
@DaveDavenport fixed

Issue #2123, opened 2025-04-15
[BUG] Rofi does not detect [133] key press when fcitx5 is running


imsuck:
### Rofi version (rofi -v)

`1.7.8` and `1.7.8-31-gf0ccf4c1 (makepkg)`

### Configuration

https://gist.github.com/imsuck/36ad7b25cee9244c42c82c53734a4e33#file-config-rasi

### Theme

https://gist.github.com/imsuck/36ad7b25cee9244c42c82c53734a4e33#file-theme-rasi

### Launch command

rofi -show drun -kb-cancel '[133]'

### Step to reproduce

1. Have fcitx5 running in the background.
2. Run `rofi -show drun -kb-cancel '[133]'`
3. Press the Super key, rofi does not close.

### Expected behavior

Rofi closes after pressing Super.

### Actual behavior

Rofi does not close.

### Additional information

Running `G_MESSAGES_DEBUG=IMDKit rofi -show keys -kb-cancel "[133]"` with fcitx5 gives:

```
Mode keys is not enabled. I have enabled it for now.
Please consider adding keys to the list of enabled modes: modes: [drun,keys].
(process:5049): IMDKit-DEBUG: 08:01:41.255: release handler 36
(process:5049): IMDKit-DEBUG: 08:01:42.004: press key 133 to xim
(process:5049): IMDKit-DEBUG: 08:01:42.008: press handler 133
(process:5049): IMDKit-DEBUG: 08:01:42.090: release key 133 to xim
(process:5049): IMDKit-DEBUG: 08:01:42.097: release handler 133
(process:5049): IMDKit-DEBUG: 08:01:42.467: press key 36 to xim
(process:5049): IMDKit-DEBUG: 08:01:42.469: press handler 36
```

And without fcitx5:
```
Mode keys is not enabled. I have enabled it for now.
Please consider adding keys to the list of enabled modes: modes: [drun,keys].
(process:13059): IMDKit-DEBUG: 08:06:25.594: release handler 36
(process:13059): IMDKit-DEBUG: 08:06:25.977: press handler 133
```

### Using wayland display server protocol

- [x] No, I don't use the wayland display server protocol

### I've checked if the issue exists in the latest stable release

- [x] Yes, I have checked the problem exists in the latest stable version

DaveDavenport:
Is super key bound do something in fcitx5?

imsuck:
[...]

PR #2124, opened 2025-04-26
Add -imdkit config to toggle imdkit at runtime

amosbird:
See discussion in #2123


DaveDavenport:
Thanks!!
maybe 'imdkit' is better then 'enable-imdkit'.. because now it is:

-enable-imdkit

and 

-no-enable-imdkit

if you f.e. want to override config file.

amosbird:
Oh, sure!

PR #2148, opened 2025-06-11
[Wayland] Estimate DPI from one detected monitor

lbonn:
This will work for single-monitor setup

Co-Authored-by: lbonn


lbonn:
@DaveDavenport lgtm

PR #2178, opened 2025-09-02
Fixed -transient-window

pijulius:
Just a quick fix re 2.0.0 release, it seems it got lost through the rewrite so added it back so -transient-window works again as before on x11 backend.


DaveDavenport:
I am so sorry I missed this getting lost in the merge.. (It was pretty big.).

Thanks for you quick MR ! Super!

PR #2186, opened 2025-09-14
feat: support `text-input-unstable-v3`

Hanssen0:
This PR implements the `text-input-unstable-v3 protocol` for the Rofi Wayland backend. This is the standard protocol for handling advanced text input, including Input Method Editors (IMEs) and virtual keyboards.

Currently, Rofi's Wayland support for complex text input (e.g., for Chinese, Japanese, and Korean) is limited. The `text-input-unstable-v3` protocol provides a modern and more robust mechanism for integrating with IMEs like `Fcitx5`.

Related issue: [[REQUEST] input method support for wayland](https://github.com/lbonn/rofi/issues/131)

This PR consists of two commits:
1. Protocol Support (`text-input-unstable-v3`).
2. A new configuration option `-layer` to specify rofi's Wayland layer.

The current & default Wayland layer is `overlay` ([[BUG] Rofi being drawn under fullscreen window](https://github.com/lbonn/rofi/issues/12)). This may prevent Rofi from being hidden by a full-screen window (though it's not reproduced on my side, with Hyprland), but it also hides the input method candidate list. Setting the layer to `top` solves the problem.

Notice: Pre-edit text rendering is not implemented. I want to keep the content of this PR as small as possible by avoiding changes to `textbox` (or related theme styles). Also, this feature seems not to be implemented on the Rofi X11 backend.

### DIFF
--- a/doc/rofi.1.markdown
+++ b/doc/rofi.1.markdown
@@ -180,6 +180,11 @@ Default:  Autodetect
 
 The X server to contact. Default is `$DISPLAY`.
 
+`-wayland-layer` *layer*
+
+On Wayland, specifies the layer where rofi is rendered. Available layers are
+`background`, `bottom`, `top`, `overlay`. The default layer is `overlay`.
+
 `-dmenu`
 
 Run **rofi** in dmenu mode. This allows for interactive scripts.
--- a/meson.build
+++ b/meson.build
@@ -318,6 +318,7 @@ if wayland_enabled
         wayland_sys_protocols_dir + '/stable/xdg-shell/xdg-shell.xml',
         wayland_sys_protocols_dir + '/unstable/primary-selection/primary-selection-unstable-v1.xml',
         wayland_sys_protocols_dir + '/unstable/keyboard-shortcuts-inhibit/keyboard-shortcuts-inhibit-unstable-v1.xml',
+        wayland_sys_protocols_dir + '/unstable/text-input/text-input-unstable-v3.xml',
         'protocols/wlr-foreign-toplevel-management-unstable-v1.xml',
         'protocols/wlr-layer-shell-unstable-v1.xml',
     )
--- a/source/wayland/display.c
+++ b/source/wayland/display.c
@@ -66,6 +66,7 @@
 #include "keyboard-shortcuts-inhibit-unstable-v1-protocol.h"
 #include "primary-selection-unstable-v1-protocol.h"
 #include "wlr-layer-shell-unstable-v1-protocol.h"
+#include "text-input-unstable-v3-protocol.h"
 
 #define wayland_output_get_dpi(output, scale, dimension)                       \
   ((output)->current.physical_##dimension > 0 && (scale) > 0                   \
@@ -113,6 +114,8 @@ static gboolean wayland_display_late_setup(void);
 static wayland_stuff wayland_;
 wayland_stuff *wayland = &wayland_;
 static const cairo_user_data_key_t wayland_cairo_surface_user_data;
+
+static const struct zwp_text_input_v3_listener text_input_listener;
 
 static void wayland_buffer_cleanup(wayland_buffer_pool *self) {
   if (!self->to_free) {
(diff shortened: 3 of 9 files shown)


source/wayland/display.c
```diff
@@ -1171,6 +1184,75 @@ static const struct wl_seat_listener wayland_seat_listener = {
...
+                                     struct zwp_text_input_v3 *text_input,
+                                     const char *text) {
+  if (text == NULL) {
+    return;
+  }
+
+  RofiViewState *state = rofi_view_get_active();
+  if (state) {
+    // rofi_view_handle_text needs a mutable string but not used
+    rofi_view_handle_text(state, (char*)text);
```
lbonn: We can change rofi_view_handle to take `const char*` probably?

meson.build
```diff
@@ -320,6 +320,7 @@ if wayland_enabled
         wayland_sys_protocols_dir + '/unstable/keyboard-shortcuts-inhibit/keyboard-shortcuts-inhibit-unstable-v1.xml',
         'protocols/wlr-foreign-toplevel-management-unstable-v1.xml',
         'protocols/wlr-layer-shell-unstable-v1.xml',
+        'protocols/text-input-unstable-v3.xml',
```
lbonn: It's been shipping with [wayland-protocols](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/tree/main/unstable/text-input) for a while.

Maybe just look in wayland_sys_protocols_dir instead of copying it into rofi?

doc/rofi.1.markdown
```diff
@@ -180,6 +180,11 @@ Default:  Autodetect
 
 The X server to contact. Default is `$DISPLAY`.
 
+`-layer` *layer*
```
lbonn: That's a cool change thanks! I guess with the variety of setups around, there is no one-size fit all layer.

But this config entry's name may be a bit short.

`-layer` does not really convey it's about wayland/layer-shell by itself and has no effect on X11.

Maybe `-wayland-layer`, though it's not great either...
@DaveDavenport any thoughts?

lbonn (commented):
Thanks a lot for working on this! I've tested it with fcitx5 and it's working good (well from what I know about input methods...)

source/wayland/display.c
DaveDavenport: Yes, it probably should be 'const', the textbox_append_text it passes to is also const.


(I've tried in the past to fix  'const' things)

DaveDavenport: fixed in rofi.

doc/rofi.1.markdown
DaveDavenport: I think `-wayland-layer` is probably the best name. It is good to be explicit.

DaveDavenport:
Thanks for your effort!

source/wayland/display.c
Hanssen0: Removed the type conversion.

meson.build
Hanssen0: Thank you for your suggestion! I'm not familiar with Wayland protocols, so I didn't notice we could directly add them from `wayland-protocols` before.

I've updated the `xml` file path.

doc/rofi.1.markdown
Hanssen0: Agree, I also struggled between `-wayland-layer` and `-layer`. At the end, I chose `-layer` because I noticed that X11 uses `-display` instead of `-x11-display` (it should be a historical reason, though), so I decided to follow the original style.

I've updated the name to `-walyland-layer` to be explicit.

lbonn:
Great! Many users have been asking for this feature :)

DreamMaoMao:
It seems there is something strange. After rofi exits, the state is not properly cleared. When it is opened next time, it still remains in the switched state from the previous time instead of the default state.


https://github.com/user-attachments/assets/a495bcbc-1ed8-48bb-a0d3-5b1ae3a90345

PR #2187, opened 2025-09-14
feat(script): `ROFI_INPUT` for custom scripts to read the user input

Hanssen0:
This pull request introduces a new environment variable, `ROFI_INPUT`, for `script` mode. This variable contains the raw text from the user's input bar, allowing for more powerful and interactive custom scripts.

Currently, script mode can only react to the final selected entry or arguments passed at startup. There is no straightforward way for a script to get the actual text the user has typed into the input bar before pressing `Enter`. This limitation prevents many powerful use cases that rely on direct user input.

An example is creating [a mode to query an AI directly](https://github.com/Hanssen0/dotfiles/blob/20158e4ed19560b18c3c62b57b8f8ac3b41f71ff/.config/rofi/scripts/ai.sh). Instead of just filtering a predefined list, the user can type a question, and the script can pass this raw input to a command-line AI tool, displaying the answer directly within Rofi.

### DIFF
--- a/doc/rofi-script.5.markdown
+++ b/doc/rofi-script.5.markdown
@@ -71,6 +71,10 @@ Environment get set when selected entry get set with the property value of the
 
 Environment get set when script sets `data` option in header.
 
+### `ROFI_INPUT`
+
+The original input string from user.
+
 ## Passing mode options
 
 Extra options, like setting the prompt, can be set by the script. Extra options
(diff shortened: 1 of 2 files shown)


DaveDavenport:
Looks good. Its a pretty low impact addition and if it helps users !

I do sometimes wonder where the 'limit' of rofi should be compared to just a UI toolkit.

Issue #2188, opened 2025-09-14
[REQUEST] Add keybinding to transpose characters (control+t)


aksr:
### Before opening a feature request

- [x] I checked the *next* branch to see if the feature has already been implemented
- [x] I searched existing reports to see if it is already requested.

### What is the user problem or growth opportunity you want to see solved?

Typing too fast is often overlooked yet very common problem.
Having a key combination to *transpose-chars* (ctrl+t) could save 4 keys:
1. Pressing delete **two times** to delete last two characters.
2. Typing **one**, then **another** character.

### How do you know that this problem exists today? Why is this important?

It isn't implemented in a release or in development repo.

### Who will benefit from it?

Emacs, readline etc. have this option — everyone will benefit.

### Rofi version (rofi -v)

2.0.0-dirty

### Configuration

It isn't needed.

DaveDavenport:
Good idea.

PR #2189, opened 2025-09-15
[Textbox] Add transpose keybinding

DaveDavenport:
Issue: #2188

### DIFF
--- a/source/widgets/textbox.c
+++ b/source/widgets/textbox.c
@@ -551,8 +551,11 @@ static void textbox_draw(widget *wid, cairo_t *draw) {
   if (tb->flags & TB_EDITABLE) {
     // We want to place the cursor based on the text shown.
     const char *text = pango_layout_get_text(tb->layout);
-    // hide the cursor, if no text is entered and hide-empty-cursor is set to true
-    if (!(tb->text[0] == '\0' && rofi_theme_get_boolean(WIDGET(tb), "hide-cursor-on-empty", FALSE) == TRUE)){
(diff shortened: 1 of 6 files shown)


source/widgets/textbox.c
```diff
@@ -823,6 +826,42 @@ static void textbox_cursor_bkspc(textbox *tb) {
...
   }
 }
+
+/**
+ * @param tb Handle to the textbox
+ *
+ * Transpose the two characters before the cursor.
+ */
+static void textbox_transpose_chars(textbox *tb) {
+  // Need to have more then 2 characters in front of the cursor.
```
lbonn, line 836: The function can also be written with early returns to avoid the nested ifs.

But that's a bit subjective, up to you :)

lbonn (approved):

source/widgets/textbox.c
DaveDavenport: yeah, I was debating this when writing it.  I suspect the compiler turns it into the same code.

DaveDavenport:
i looks like 🖖🏻 is not handled correctly by glib, so transpose screws it up.
I am not sure what is the best way to fix this, without re-inventing utf-8 handling code.

source/widgets/textbox.c
lbonn: Yes it's not about the generated code but readability. But it's just a matter of taste and this function is small anyway :)

lbonn (approved):

PR #2196, opened 2025-09-20
add switch-mode option to the script mode

tmoron:
Added an option to switch modes from a script.
Usage : 
```
echo -e "\0switch-mode\x1fdrun"
```
this switches to the drun mode

### DIFF
--- a/source/modes/script.c
+++ b/source/modes/script.c
@@ -77,7 +77,11 @@ typedef struct {
   gboolean keep_filter;
 
   gboolean use_hot_keys;
+
+  gboolean switch_mode;
 } ScriptModePrivateData;
+
+extern unsigned int curr_mode;
 
 /**
  * Shared function between DMENU and Script mode.
@@ -129,6 +133,18 @@ void dmenuscript_parse_entry_extras(G_GNUC_UNUSED Mode *sw,
  * End of shared functions.
  */
 
+static void script_switch_mode(Mode *sw, char *value) {
+  int index = mode_lookup(value);
+  ScriptModePrivateData *pd = (ScriptModePrivateData *)sw->private_data;
+
+  if (index >= 0)
+  {
+    pd->switch_mode = index;
+  } else {
+    g_warning("Mode \"%s\" not found, please make sure it's enabled.\n", value);
+  }
+}
+
 static void parse_header_entry(Mode *sw, char *line, ssize_t length) {
   ScriptModePrivateData *pd = (ScriptModePrivateData *)sw->private_data;
   ssize_t length_key = 0; // strlen ( line );
@@ -172,6 +188,8 @@ static void parse_header_entry(Mode *sw, char *line, ssize_t length) {
         g_warning("Failed to parse: '%s'", value);
         rofi_clear_error_messages();
       }
+    } else if (strcasecmp(line, "switch-mode") == 0) {
+      script_switch_mode(sw, value);
     }
   }
 }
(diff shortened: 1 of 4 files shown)


DaveDavenport:
What is the use-case for this patch?

tmoron:
I want to create a kind of submenu system in the launcher, where different prefixes trigger different modes or actions. For example, typing d would switch to drun, typing c would switch to calc, and other prefixes could be used to start Firefox with a search or open a bookmark.

source/modes/script.c
```diff
@@ -77,8 +77,12 @@ typedef struct {
   gboolean keep_filter;
 
   gboolean use_hot_keys;
+
+  int16_t switch_mode;
```
DaveDavenport: Why is this an int16_t  and not just an int (as mode_lookup returns)?

tmoron: I wanted to use the smallest type possible and since mod ids seem to conflict with ModeMode values after 999, I thought it was the limit. But yeah you're right, using an int is probably safer.

DaveDavenport: I should probably enforce the 999 limit somewhere.. Given how rofi grown :-P

tmoron: do you want me to change the type to an int ?

DaveDavenport: Yes, I think that is preferred.

source/modes/script.c
```diff
@@ -77,8 +77,12 @@ typedef struct {
...
 } ScriptModePrivateData;
 
+extern unsigned int curr_mode;
```
DaveDavenport: I am not a big fan of  this 'bypass' of the mode API. 
I need to remember exactly how it works,  Maybe we can use MENU_QUICK_SWITCH  structure.

I think if you return the mode number from the _result function it should switch to that mode..

DaveDavenport: See process_result in rofi.c

tmoron: yes, returning the id of the mode does switch to that mode,that's what I had initially. but this method doesn't clear the input and the next mode starts with the previous mode's input. So if I start drun with "d" in script, drun mode will be started with a d already entered. I avoided adding a clear in the default switch to prevent breaking other behavior.

DaveDavenport: free and set *input to NULL?

DaveDavenport: image
That should clear it.

tmoron: oh, I didn't notice that, yes that's a much better approach. i'll make a new commit

DaveDavenport: great!   I had to look, as it been a long time since I touched this part of the code.

source/modes/script.c
```diff
@@ -335,6 +351,7 @@ static ModeMode script_mode_result(Mode *sw, int mretv, char **input,
   // store them as they might be different on next executor and reset.
   gboolean keep_filter = rmpd->keep_filter;
   gboolean keep_selection = rmpd->keep_selection;
+  RofiViewState *state = rofi_view_get_active();
```
DaveDavenport: I think this one is not needed anymore.

source/modes/script.c
```diff
@@ -377,6 +394,14 @@ static ModeMode script_mode_result(Mode *sw, int mretv, char **input,
     }
   }
 
+  if (state && rmpd->switch_mode >= 0) {
```
DaveDavenport: state check is not needed anymore?

tmoron: yep, my bad sorry

DaveDavenport:
Thanks.

tmoron:
thank you !
btw, I love this project, it's awesome !

PR #2201, opened 2025-09-24
filebrowser: sort based on natural ordering

uninsane:
filebrowser: sort entries based on natural ordering.

previous behavior: `rofi -show filebrowser` would order entries by byte order, for example:
- `Music/`
- `Photos/`
- `Videos/`
- `backup/`
- `source/`

the new sort order considers these case-insensitively:
- `backup/`
- `Music/`
- `Photos/`
- `source/`
- `Videos/`

it also orders numbers naturally, e.g. `project-v2` before `project-v10`. [utf8_collate_key_for_filename](https://docs.gtk.org/glib/func.utf8_collate_key_for_filename.html) docs for complete reference.

### DIFF
--- a/source/modes/filebrowser.c
+++ b/source/modes/filebrowser.c
@@ -97,7 +97,9 @@ typedef struct {
   uint32_t icon_fetch_size;
   guint icon_fetch_scale;
   gboolean link;
+  /** {time,collate_key} only valid for FB_SORT_{TIME,NAME} respectively */
   time_t time;
+  char *collate_key;
 } FBFile;
 
 typedef struct {
@@ -132,6 +134,7 @@ static void free_list(FileBrowserModePrivateData *pd) {
     FBFile *fb = &(pd->array[i]);
     g_free(fb->name);
     g_free(fb->path);
+    g_free(fb->collate_key);
   }
   g_free(pd->array);
   pd->array = NULL;
@@ -150,7 +153,7 @@ static gint compare_name(gconstpointer a, gconstpointer b,
     return fa->type - fb->type;
   }
 
-  return g_strcmp0(fa->name, fb->name);
+  return g_strcmp0(fa->collate_key, fb->collate_key);
 }
 
 static gint compare_time(gconstpointer a, gconstpointer b,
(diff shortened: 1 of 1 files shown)


source/modes/filebrowser.c
```diff
@@ -150,7 +150,14 @@ static gint compare_name(gconstpointer a, gconstpointer b,
     return fa->type - fb->type;
   }
 
-  return g_strcmp0(fa->name, fb->name);
+  const gchar *fa_key = g_utf8_collate_key_for_filename(fa->name, -1);
+  const gchar *fb_key = g_utf8_collate_key_for_filename(fb->name, -1);
+
+  gint order = g_strcmp0(fa_key, fb_key);
+
+  g_free(fa_key);
```
lbonn: It allocates and frees for every of the n*log(n) comparisons which looks wasteful.

Not sure if it has a noticeable effect on large directories but it's worth checking.

DaveDavenport: Might be worth it to store the collate key on loading?  its extra memory usage, but avoids a lot of memory allocations/frees when sorting.

uninsane: updated the patch to store the collate key as part of `get_file_browser`. took inspiration from how we manage the `time` field and hid the collate key assignment behind an `if (sorting_method == FB_SORT_NAME)` guard, so it should have minimal impact however the user's sorting.

let me know what you think?

source/modes/filebrowser.c
```diff
@@ -254,6 +261,7 @@ static void get_file_browser(Mode *sw) {
         pd->array[pd->array_length].icon_fetch_scale = 0;
         pd->array[pd->array_length].link = FALSE;
         pd->array[pd->array_length].time = -1;
+        pd->array[pd->array_length].collate_key = g_strdup("");
```
DaveDavenport: Why not set to NULL? avoid the alloc. 
g_free/g_strcmp0 will accept NULL and do nothing.

uninsane: TIL! applied

lbonn (approved):

PR #2202, opened 2025-09-24
[Theme] add alpha override support to references and hex colors.

auregosu:
this adds support for the syntax `@color / 50%;` and `#123456 / 50%` to override the alpha of a given color.
- extended grammar rules for `t_property_element` and `t_property_color`
- added `gboolean` member to link `union` to differentiate modified links from normal links
- added definition of `rofi_theme_resolve_modified_link()`
- added implementation of modified link resolving in `rofi_theme_parse_process_links_int()`

https://github.com/davatorium/rofi/discussions/2200

### DIFF
--- a/lexer/theme-parser.y
+++ b/lexer/theme-parser.y
@@ -570,6 +570,13 @@ t_property_element
         $$ = rofi_theme_property_create ( P_LINK );
         $$->value.link.name = $1;
     }
+ /** @color_reference / 0-100% */
+|   T_LINK T_FORWARD_SLASH t_property_color_value_unit {
+	$$ = rofi_theme_property_create ( P_COLOR );
+	$$->value.link.name = $1;
+	$$->value.color.alpha = $3;
+	$$->value.link.modified = TRUE;
+    }
 |   T_BOOLEAN {
         $$ = rofi_theme_property_create ( P_BOOLEAN );
         $$->value.b = $1;
@@ -999,8 +1006,10 @@ t_property_color
     $$.alpha = $8;
 }
 /** Hex colors parsed by lexer. */
-| T_COLOR {
+| T_COLOR t_property_color_opt_alpha_ws {
     $$ = $1;
+	if ($2 != 1.0)
+		$$.alpha = $2;
 }
 | T_COLOR_TRANSPARENT {
     $$.alpha = 0.0;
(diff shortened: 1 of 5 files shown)


lexer/theme-parser.y
```diff
@@ -999,8 +1006,9 @@ t_property_color
     $$.alpha = $8;
 }
 /** Hex colors parsed by lexer. */
-| T_COLOR {
+| T_COLOR t_property_color_opt_alpha_ws {
     $$ = $1;
+    $$.alpha  = $2;
```
DaveDavenport: This overrides the alpha that is (possibly) already set.

maybe a `*=`  instead of `=`?

source/theme.c
```diff
@@ -1592,6 +1592,16 @@ static void rofi_theme_parse_process_links_int(ThemeWidget *wid) {
           }
         }
       }
+      if (pv->type == P_COLOR && pv->value.link.modified == 1) {
+        if (pv->value.link.ref == NULL) {
+			rofi_theme_resolve_modified_link(pv);
+			if (pv->value.link.ref) {
+				ThemeColor referenced_color = pv->value.link.ref->value.color;
```
DaveDavenport, line 1599: is it guaranteed that the 'link'  points to a color?

auregosu: i'll add a check for the type of `pv->value.link.ref` on line 1598. should there be a warning if it's otherwise?

DaveDavenport: A g_debug should be fine and the color should not be set. 
(I think this is normal behaviour if you try to set color to f.e. a boolean..   the color is not set, and 'Theme entry: ... unset"  is printed in debug.)

lexer/theme-parser.y
auregosu: i made it analogous to the already existing behavior for named colors on line 1019:
```
| T_COLOR_NAME t_property_color_opt_alpha_ws {
    $$ = $1;
    $$.alpha = $2;
}
```
it's probably fine to leave it this way.

auregosu:
sorry for force pushing yet again. i corrected the alpha overriding of hex colors so it only happens when `t_property_color_opt_alpha_ws` is set. i also fixed a naive mix-up of string assignment and copying that lead to other errors.

PR #2267, opened 2026-02-15
[WIP] [perf] Add fast desktop entry parser, ThorVG icons, and sync loading

markg85:
Performance improvements:
- AVX2 SIMD-optimized desktop entry parser (~32 bytes/cycle)
- ThorVG for icon rendering (~50μs per icon vs ms with gdk-pixbuf)
- Pre-built icon path cache for O(1) lookups
- Removed async icon loading (threading overhead exceeds work at ~50μs)

Note: Optimized for SSD/NVMe storage

----

Hi,

I've been using rofi for a while and i like it. However, i noticed that icons sometimes took a fraction of time to load (showing an empty spot, then the icon popping in).

I would've never debugged this as i don't like c and don't like GTK either and would've just used it as is. However, these days we have AI and i happen to be interested in performance optimizations! That makes it bearable to dive into a repository and see how it can be improved. for those wondering about the AI model i used. I used GLM 5 with Claude Code.

Turns out it could be improved truly massively. Which this single commit fork is doing!
And yes, this was in release mode with debug symbols (-g flag).
```
  ┌────────┬────────────┬────────┬──────────────┐
  │ Format │ gdk-pixbuf │ thorvg │   Speedup    │
  ├────────┼────────────┼────────┼──────────────┤
  │ SVG    │ 4310 µs    │ 212 µs │ 20.3x faster │
  ├────────┼────────────┼────────┼──────────────┤
  │ PNG    │ 4249 µs    │ 142 µs │ 29.9x faster │
  └────────┴────────────┴────────┴──────────────┘

  ┌───────────────┬────────────┬───────────────────────┐
  │    Method     │    Time    │       Per Icon        │
  ├───────────────┼────────────┼───────────────────────┤
  │ nk_xdg lookup │ 550.038 ms │ 1.95 ms               │
  ├───────────────┼────────────┼───────────────────────┤
  │ cached lookup │ 0.036 ms   │ 0.000128 ms (0.13 µs) │
  ├───────────────┼────────────┼───────────────────────┤
  │ Speedup       │            │ 15,412x faster!       │
  └───────────────┴────────────┴───────────────────────┘
  ```
  
Moving away from gdk-pixbuf to [ThorVG](https://github.com/thorvg/thorvg) is a direct massive speedup for loading images. I was honestly surprised by this, it's a rather massive difference. That alone makes an individual icon load within a mean time of ~0.5ms on my pc. It does differ per png icon and complexity. What you see in the benchmark is just the first icon it found which was from alacritty. ThorVG supports loading of a few formats including png so with that in mind i just removed gdk-pixbuf entirely.

This single optimization wasn't enough though. I wanted the top 10 images to load with 1ms (all of them! so 0.1ms per image). Turns out that the icon lookup path was quite heavy too so i optimized that too. Essentially by building a hashmap as lookup.

With these two changes (and a lot of micro optimizations like reducing allocations, custom desktop file parsing, avx2, etc...) the loading was now so incredibly fast that the use of async was turning into a UI glitch. Or put differently, you could still see a result where the icon wasn't loaded but popped in the next frame. I solved this by simply removing the async path and going for sync. I do realize that this change can potentially be bad too as it might block ui based in icon loading. So for the HDD users, probably don't use this patch. But if you use an SSD or NVMe then you're fast enough to not notice the difference. To me at least the UI feels snappier then it ever did (and it was already quite good!).

Lastly, i also removed the caching mechanism completely. Having a fast cache is great! Having no cache that is even faster is just better :) In my case my rofi top 10 was takes ~300ms to load (hence a short hiccup was visible) to now on my pc just ~1.3ms (that's the list and ui for the list).

For the Rofi project. I doubt these changes are all going to be merged as-is. I'd recommend to take this as inspiration and cherry pick parts of if for a future version. Like changing to ThorVG is essentially a free speed boost.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,37 @@
 <a href="https://repology.org/metapackage/rofi/versions"><img src="https://repology.org/badge/tiny-repos/rofi.svg"></a>
 </p>
 
+## Performance Improvements Over Upstream
+
+This fork includes significant performance optimizations focused on startup speed and responsiveness:
+
+### Fast Desktop Entry Parser (DRUN mode)
+- **AVX2 SIMD optimization**: Uses 256-bit vector instructions to scan for `[Desktop Entry]` sections - processes 32 bytes per cycle
+- **Single-pass parsing**: Minimal memory allocations with direct string comparisons
+- **8-byte load comparisons**: Scalar fallback using 64-bit loads for fast scanning on non-AVX2 systems
+- **Runtime CPU detection**: Automatically uses AVX2 when available, falls back gracefully
+
+### ThorVG Image Rendering
+- **Replaced gdk-pixbuf with ThorVG**: Icon loading reduced from milliseconds to **~50μs** per icon
+- **Broad format support**: SVG, SVGZ, PNG, JPG, JPEG, WEBP, TVG
+- **Hardware-optimized**: ThorVG is designed for embedded and high-performance rendering
+
+### Fast Icon Path Cache
+- **Pre-built hash table**: Icon paths indexed at startup for **O(1) lookups** instead of O(n) directory scans
+- **No filesystem traversal**: Eliminates repeated directory scans during icon lookup
+- **Size-aware caching**: Stores icons by name and size for optimal resolution matching
+
+### Synchronous Icon Loading (Async Removed)
+- **Icon loading is now ~50μs**: Fast enough that async threading overhead exceeds the actual work
+- **No thread overhead**: Eliminates mutex contention, context switches, and callback complexity
+- **Immediate availability**: Icons are ready when requested, no waiting for worker threads
+
+> **⚠️ Storage Requirement**: These performance gains assume **SSD/NVMe storage**. On traditional HDDs, image loading may take longer and async could theoretically help. This fork is optimized for modern storage.
+
+**Result**: Significantly faster startup and instant icon display on systems with SSD/NVMe storage.
+
+---
+
 **Please match the documentation and scripts to the version of rofi used**
 
 - [next version](https://github.com/davatorium/rofi)
--- a/meson.build
+++ b/meson.build
@@ -62,7 +62,7 @@ deps = [
     dependency('pango'),
     dependency('pangocairo'),
     dependency('xkbcommon'),
-    dependency('gdk-pixbuf-2.0'),
+    dependency('thorvg-1'),
     dep_lm,
 ]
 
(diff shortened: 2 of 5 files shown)


DaveDavenport:
interesting that thorvg is so much faster loading images. That could be interesting even with the async loading.

If I test it on my machine here (i7-13700H) it pops up at the same speed (~0.2s). But on a fresh loads indeed feel slightly faster as the icons pops in slightly quicker (compared to rofi on redraw-fixes branch). 
However on filebrowser on an image directory with larger images it is significantly slower (as expected because it being 'sync' now).

The drun parser optimization here saves around 1ms (4->3ms), the cache (0.12ms) still out performs it with significant margin.  Therefor I am not sure I completely understand this remark:

>Lastly, i also removed the caching mechanism completely. Having a fast cache is great! Having no cache that is even faster is just better :) In my case my rofi top 10 was takes ~300ms to load (hence a short hiccup was visible) to now on my pc just ~1.3ms (that's the list and ui for the list).


To bad it is one massive patch, with hardcoded assumptions in there. Not sure this will work well for lot of people. 
Speed improvements are always welcome, but I am afraid (as you say) this is not directly usable. 


(on a side note, your fork does not build, the benchmark-drun.c is missing)

markg85:
Hi Dave,

I'd consider it a suggestion at potential improvements. Not a PR to accept :) (which you didn't as you closed it, good!)

Now you can consider it motivation for yourself, that's awesome! But I'm also able and willing to provide more fine-grained patches if you like the idea of these optimizations.

I might have benchmarked this a little too liberate and drawn conclusions from it that might not have been fair in hindsight. My non-cache version was outperforming the cache but i admit that i forgot to clear ram after each benchmark. Which i obviously should've done as it can remain in hot cache if you don't do that and if it fits ram. So that's a mistake on my end.

>  Therefor I am not sure I completely understand this remark
Well the remark was made with the above in mind. If that's not true then the remark is wrong too :)

Do you want me to provide individual patches? I can imagine the following ones:
- replace gdk-pixbuf with thorvg
- optimize desktop index parser with a custom parser
- icon lookup optimizations

DaveDavenport:
We currently do not (knowingly) accept AI written code into rofi.

Beside that, there are to many issues (not all mentioned above) with the code to be safely merged in.

I also been (on multiple machines, with varying specs)  unable to confirm the speedups in the use of rofi. Its either marginally faster or significantly slower. This is not due to hot/cold filesytem cache.

I currently do not have a lot of time to work on rofi, and evaluating this already took most of the time I had.

markg85:
The never ending fear of accepting AI cooked patches ;)
It's your time and stance which i totally accept and appreciate! But at the same time a feeling of "well, your loss" does creep up.

Don't forget, this is human steered and verified but ai written. AI would not come as far in these optimizations if i just tell it to "optimize project y, pull out icon loading, profile it and make it 100x faster" it won't get anywhere near the speeds i ended up measuring. It needs a lot of steering and correcting before it can do in hours where i would've spend weeks if i had done it manually.

DaveDavenport:
Its clear you did not read my reply beyond the first ~sentence~ remark.

I did evaluate the patch and found it did not live up to the claims made in my testing and often produced a slower result. Beside this the "quality" of the patch is to poor to be merged as-is (f.e. hardcoded paths that will not make it work well on systems, hardcoded list of supported image formats, missing formats etc.).
This took a fair bit of time to do.
The parts that might be useful, would require to much effort to get right then I currently have.


I did not only reject it solely of it being generated by AI.

PR #2272, opened 2026-03-08
wayland: implement click-to-exit support using fullscreen capture surface

### LINKED ISSUE #2158: [REQUEST] Ability to close rofi on outside click.
### Before opening a feature request

- [x] I checked the *next* branch to see if the feature has already been implemented
- [x] I searched existing reports to see if it is already requested.

### What is the user problem or growth opportunity you want to see solved?

Currently, the only way to close rofi is by either selecting something from the menu or pressing Esc. 

I'm a very big noob to this, but is it not possible to just add a "button" that fills up the entire background of rofi to close the window? Imagine clicking anywhere "outside" the rofi window to close rofi. its not like you can click outside of rofi when its open anyway. (unless you use a script)

Again, I only got into linux like 3 days ago so bear with me haha!

### How do you know that this problem exists today? Why is this important?

its annoying having to reach for the escape button every time.

### Who will benefit from it?

probably everyone

### Rofi version (rofi -v)

Version: 1.7.9.1-dirty 

### Configuration

N/A

### Additional information

_No response_

benny-e:
This PR implements native click-to-exit support for rofi's Wayland backend.

On Wayland, rofi cannot receive pointer events outside its surface, so
click-to-exit does not work when the surface only matches the visible
menu size. This patch addresses that by using a fullscreen transparent
surface when click-to-exit is enabled.

The visible menu keeps its normal size and theme-driven placement
inside that fullscreen surface. Pointer events are compared against the
visible menu rectangle so inside clicks continue through the normal
widget path, while outside clicks cancel the active view.

When click-to-exit is disabled, existing Wayland behavior remains
unchanged.

Fixes #2158

Tested with:
- `rofi -dmenu -no-config -no-click-to-exit`
- `rofi -dmenu -no-config -click-to-exit`
- `rofi -show drun -no-click-to-exit`
- `rofi -show drun -click-to-exit`

Also tested with non-centered placement to verify that theme-driven
positioning is preserved in capture mode.

Tested on Hyprland and Sway with single and multi-monitor setups. Outside clicks close as expected, including clicking on another monitor.

### DIFF
--- a/include/view-internal.h
+++ b/include/view-internal.h
@@ -195,6 +195,9 @@ typedef struct _view_proxy {
   void (*get_size)(struct RofiViewState *state, gint *width, gint *height);
 
   void (*pool_refresh)(void);
+
+  void (*get_menu_rect)(int *x, int *y, int *w, int *h);
+
 } view_proxy;
 
 typedef struct {
--- a/include/view.h
+++ b/include/view.h
@@ -29,8 +29,8 @@
 #define ROFI_VIEW_H
 
 #include "mode.h"
-#include "widgets/widget.h"
 #include "widgets/textbox.h"
+#include "widgets/widget.h"
 #include <pango/pango.h>
 #ifdef ENABLE_XCB
 #include <xcb/xcb.h>
@@ -213,10 +213,10 @@ void rofi_view_free(RofiViewState *state);
 RofiViewState *rofi_view_get_active(void);
 
 /**
-  * Get the current active textbox with the user input.
-  *
-  * @returns the active textbox or NULL
-  */
+ * Get the current active textbox with the user input.
+ *
+ * @returns the active textbox or NULL
+ */
 textbox *rofi_view_get_active_text(void);
 
 /**
@@ -372,7 +372,10 @@ void rofi_view_ping_mouse(RofiViewState *state);
 void rofi_view_set_window_title(const char *title);
 void rofi_view_pool_refresh(void);
 
+void rofi_view_get_menu_rect(int *x, int *y, int *w, int *h);
+
 void rofi_view_set_cursor(RofiCursorType type);
+
 
 /**
  * Input history
@@ -407,6 +410,9 @@ void rofi_view_ellipsize_listview(RofiViewState *state,
  * main_window
  */
 gboolean rofi_set_im_window_pos(int new_x, int new_y);
+
+// Cancel the current view
+void rofi_view_cancel(RofiViewState *state);
 
 /**
  * @param wid to test.
--- a/source/wayland/display.c
+++ b/source/wayland/display.c
@@ -117,6 +117,8 @@ static const cairo_user_data_key_t wayland_cairo_surface_user_data;
 
 static const struct zwp_text_input_v3_listener text_input_listener;
 
+static const struct _view_proxy *wayland_display_view_proxy(void);
+
 static void wayland_buffer_cleanup(wayland_buffer_pool *self) {
   if (!self->to_free) {
     return;
@@ -583,8 +585,27 @@ static void wayland_pointer_send_events(wayland_seat *self) {
     return;
   }
 
+  int menu_x = 0, menu_y = 0, menu_w = 0, menu_h = 0;
+
+  gboolean capture = config.click_to_exit;
+
+  if (capture) {
+    rofi_view_get_menu_rect(&menu_x, &menu_y, &menu_w, &menu_h);
+    if (menu_w <= 0 || menu_h <= 0) {
+      capture = FALSE;
+    }
+  }
+
   if (self->motion.x > -1 || self->motion.y > -1) {
-    rofi_view_handle_mouse_motion(state, self->motion.x, self->motion.y,
+    int motion_x = self->motion.x;
+    int motion_y = self->motion.y;
+
+    if (capture) {
+      motion_x -= menu_x;
+      motion_y -= menu_y;
+    }
+
+    rofi_view_handle_mouse_motion(state, motion_x, motion_y,
                                   config.hover_select);
     self->motion.x = -1;
     self->motion.y = -1;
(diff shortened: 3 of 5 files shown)


source/wayland/view.c
```diff
@@ -340,7 +486,6 @@ static void wayland_rofi_view_update(RofiViewState *state, gboolean qr) {
  *
  * Check if a finalize function is set, and if sets executes it.
  */
-void process_result(RofiViewState *state);
```
lbonn: You've moved the prototype definition but not the doxygen description. It's the source of the CI failure

benny-e: I have removed the stray comment.

source/wayland/view.c
lbonn: This function got quite big, I'd suggest structuring it this way:

```
static void wayland_rofi_view_window_update_size(RofiViewState *state) {
  if (WlState.capture_click_outside) {
    window_update_size_with_outside_click(state);
    return;
  }
  // simpler no outside click version
}
```

benny-e: Yeah good call. I have separated the logic for outside click to keep it smaller

lbonn:
@benny-e thanks a lot for working on this. I haven't been able to test it yet but will try to do so this week

source/wayland/view.c
```diff
@@ -139,19 +152,113 @@ static int rofi_get_offset_px(RofiViewState *state, RofiOrientation ori) {
...
 }
 
+void wayland_rofi_view_cancel(RofiViewState *state) {
+  if (state == NULL) {
+    return;
+  }
+
+  state->retv = MENU_CANCEL;
+  state->quit = TRUE;
+}
```
lbonn: It would make sense to just put this in view.c, next to `rofi_view_restart` (`rofi_view_quit`)

source/wayland/view.c
```diff
@@ -211,15 +318,25 @@ static void wayland_rofi_view_queue_redraw(void) {
...
+
+  /* Reset Wayland-local click-capture state for this view instance. */
+  WlState.capture_click_outside = config.click_to_exit ? TRUE : FALSE;
+  WlState.surface_width = 0;
+  WlState.surface_height = 0;
+  WlState.menu_width = 0;
+  WlState.menu_height = 0;
+  WlState.menu_x = 0;
+  WlState.menu_y = 0;
+
```
lbonn: You can set all of these 0s inside the static struct definition around line 100

source/wayland/view.c
```diff
@@ -211,15 +318,25 @@ static void wayland_rofi_view_queue_redraw(void) {
...
   TICK_N("create cairo surface");
-  // TODO should we update the drawable each time?
   PangoContext *p = pango_context_new();
   pango_context_set_font_map(p, pango_cairo_font_map_get_default());
   TICK_N("pango cairo font setup");
 
   WlState.flags = menu_flags;
+
+  /* Reset Wayland-local click-capture state for this view instance. */
+  WlState.capture_click_outside = config.click_to_exit ? TRUE : FALSE;
```
lbonn: I don't think you need to re-define this. Just use `config.click_to_exit` instead of `WlState.capture_click_outside` everywhere

source/wayland/view.c
```diff
@@ -139,19 +152,113 @@ static int rofi_get_offset_px(RofiViewState *state, RofiOrientation ori) {
...
+  case WL_CENTER:
+  default:
+    WlState.menu_x =
+        (WlState.surface_width - WlState.menu_width) / 2 + offset_x;
+    WlState.menu_y =
+        (WlState.surface_height - WlState.menu_height) / 2 + offset_y;
+    break;
+  }
+
+  display_set_surface_dimensions(WlState.surface_width, WlState.surface_height,
```
lbonn, line 240: One interesting side-effect of these changes is multi monitor setup:

- two monitors side by side
- rofi's window set to > screen size
- starting rofi on the left monitor

With -no-click-to-exit, the window overflows to the right's monitor and it is possible to click from there.

With -click-to-exit (or default), the overflowing part is just clipped. And click to exit does not work when clicking on other monitors (this would require setting up more surfaces probably...)

Not really sure how big of a deal all this is but it's worth being aware of the limitations. Maybe as a code comment somewhere.

lbonn:
Apart from the outside-monitor parts being clipped, it is working as advertised :)

source/wayland/view.c
benny-e: Moved to view.c

source/wayland/view.c
benny-e: Moved initialization to the static definition

source/wayland/view.c
benny-e: Removed the redefine and used config.click_to_exit

source/wayland/view.c
benny-e: I was able to replicate this when the rofi menu is larger than the output. I added a comment to document the limitation. Supporting that case would probably require more surfaces, so I left it out of the scope for this fix.

source/wayland/display.c
```diff
@@ -22,6 +22,8 @@
  *   SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+#include <stdio.h>
```
lbonn: You can probably remove this

source/wayland/display.c
```diff
@@ -604,17 +622,43 @@ static void wayland_pointer_send_events(wayland_seat *self) {
...
+      inside = self->button.x >= menu_x && self->button.x < menu_x + menu_w &&
+               self->button.y >= menu_y && self->button.y < menu_y + menu_h;
+    }
+
     if (self->button.pressed) {
-      rofi_view_handle_mouse_motion(state, self->button.x, self->button.y,
-                                    FALSE);
+      if (capture && !inside) {
+
+        wayland_rofi_view_cancel(state);
```
lbonn: To rename after the change (CI is failing)

include/view.h
```diff
@@ -408,6 +408,13 @@ void rofi_view_ellipsize_listview(RofiViewState *state,
  */
 gboolean rofi_set_im_window_pos(int new_x, int new_y);
 
+// Cancel the current view
+void rofi_view_cancel(RofiViewState *state);
+
+// Wayland click-to-exit helper
+gboolean wayland_rofi_view_capture_click_outside_enabled(void);
```
lbonn:
```suggestion
```

Same as the previous comment, this whole function can be removed and config.click_to_exit used in wayland/display.c

include/view.h
```diff
@@ -408,6 +408,13 @@ void rofi_view_ellipsize_listview(RofiViewState *state,
...
+void wayland_rofi_view_get_menu_rect(int *x, int *y, int *w, int *h);
```
lbonn: Until now we avoided having `wayland_xx` function declarations in view.h.

Could you please define `get_menu_rect` in _view_proxy instead, like we do with other methods?

benny-e:
removed the stray stdio.h include
removed the Wayland-specific declarations from view.h
moved get_menu_rect behind _view_proxy

include/view.h
```diff
@@ -408,6 +408,9 @@ void rofi_view_ellipsize_listview(RofiViewState *state,
  */
 gboolean rofi_set_im_window_pos(int new_x, int new_y);
 
+// Cancel the current view
+void rofi_view_cancel(RofiViewState *state);
```
lbonn, line 415: It looks like the implementation disappeared from view.c

include/view-internal.h
```diff
@@ -195,6 +195,10 @@ typedef struct _view_proxy {
   void (*get_size)(struct RofiViewState *state, gint *width, gint *height);
 
   void (*pool_refresh)(void);
+
+  gboolean (*capture_click_outside_enabled)(void);
```
lbonn: Sorry to repeat again but this method is not needed.
config is accessible globally.

https://github.com/davatorium/rofi/pull/2272#discussion_r2923011460

Line 591 of wayland/display.c, just check for config.click_to_exit

benny-e: Removed the helper and now check config.click_to_exit directly in wayland/display.c

include/view.h
benny-e: Ensured rofi_view_cancel() is implemented in view.c

lbonn:
@benny-e thanks, I think it's close to a mergeable state.

Do you mind if I squash all these changes in one commit with a meaningful message?
You can also do it yourself if you prefer.

benny-e:
> @benny-e thanks, I think it's close to a mergeable state.
> 
> 
> 
> Do you mind if I squash all these changes in one commit with a meaningful message?
> 
> You can also do it yourself if you prefer.


Great! I don't mind, feel free to squash

Issue #2165, opened 2025-07-28
[Wayland] Rofi icons flickers on startup


DaveDavenport:
### Discussed in https://github.com/davatorium/rofi/discussions/2160

<div type='discussions-op-text'>

<sup>Originally posted by **mazksr** July 17, 2025</sup>
https://github.com/user-attachments/assets/4d1920f4-f4c5-4519-af11-8bd327a83738

Happens on every theme with icons. Even the built-in ones.</div>

Version: pre 2.0 release after merge.

Zebra2711:
It seems rofi loads first, then the icon loads. If the icon is loaded, then it will not flicker anymore. Does rofi have an icon cache?

https://github.com/user-attachments/assets/43a998fe-0426-4d1e-bc52-50e2bc3a9930

DaveDavenport:
the explanation  for the flicker is in the linked topic. Its still on my todo to fix.

virtuallyunknown:
I am having a similar issue to this and to #2206, except that it doesn't flicker while it type, but the icons load noticeably slower than the rest.

https://github.com/user-attachments/assets/3197c90c-0fc0-4863-9cac-2634933c866f

System is Arch with Niri compositor and nvidia-open drivers, wayland. My config is super simple as this is the first time I am trying out rofi.

```rasi
configuration {
  modes: [ combi ];
  combi-modes: [  drun ];
  show-icons: true;
  font: "JetBrains Mono Nerd Font 12";
  icon-theme: "Tela-circle-pink-dark";
}
```

If requested, I can provide more technical information. All the best!

DaveDavenport:
@virtuallyunknown  this is intended behaviour,  icons are loaded asynchronous from the rest. This is done to make sure the loading does not slow down rofi..    I will mark your post as off-topic as I do not want to 'polute' this issue with other stuff.
Please read the guidelines!!!!!!

virtuallyunknown:
I apologize for going off-topic, but trust me it wasn't for the lack of research on my part. I've searched online for probably an hour for issues related to icons, and I could not find anything.

Only now that you mention this I was able to find relevant information by searching the codebase with the keyword `asynchronous`, and the only mention of this problem is in [the 2.0 release notes](https://github.com/search?q=repo%3Adavatorium%2Frofi%20asynchronous&type=code), which I hope you'll agree is not the most intuitive place to look for that. If there is a separate guidelines page, I am not sure where it's located.

[...]

PR #2277, opened 2026-03-23
[WIP] feat: add icon cache

Zebra2711:
- [x] This PR targets the `next` branch and not `master`
- [x] If your PR is a work in progress, include [WIP] in its title
- [x] Its commits' summaries are reasonably descriptive
- [x] You've described what this PR addresses below

related to #2165

save cache in `~/.cache/rofi-icon-cache/<hash>.rofi`

### DIFF
--- a/source/rofi-icon-fetcher.c
+++ b/source/rofi-icon-fetcher.c
@@ -51,7 +51,6 @@
 #include "helper.h"
 #include <gdk-pixbuf/gdk-pixbuf.h>
 
-
 /** Desktop entry specifying the thumbnailer. */
 #define THUMBNAILER_ENTRY_GROUP "Thumbnailer Entry"
 /** Extension used for the thumbnailer. */
--- a/source/rofi-icon-fetcher.c
+++ b/source/rofi-icon-fetcher.c
@@ -72,19 +72,28 @@ typedef struct {
 } IconCacheHeader;
 
 static char *icon_cache_path(const char *icon_path, int w, int h) {
-  uint32_t hash = 5381;
-  for (const char *p = icon_path; *p; p++)
+  uint32_t hash = 5381; // djb2
+  for (const char *p = icon_path; *p; p++) {
     hash = ((hash << 5) + hash) ^ (uint8_t)*p;
+  }
   hash ^= (uint32_t)(w * 10007u ^ h * 100003u);
-  return g_strdup_printf("%s/" ICON_CACHE_DIR "/%08x.rofi", g_get_user_cache_dir(), hash);
+
+  char fname[14]; // 8 hex + ".rofi" + "\0"
+  g_snprintf(fname, 14, "%08x.rofi", hash);
+  char *path = g_build_filename(g_get_user_cache_dir(), ICON_CACHE_DIR, fname, NULL);
+  return path;
 }
 
 static cairo_surface_t *icon_cache_load(const char *cache_key, const char *src_path, int w, int h, char **resolved_path_out) {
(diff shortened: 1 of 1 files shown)


source/rofi-icon-fetcher.c
```diff
@@ -50,10 +51,108 @@
...
+  int32_t  width;
+  int32_t  height;
+  int64_t  src_mtime;
+  int64_t  src_size;
+  char     src_path[ICON_CACHE_PATH_MAX];
+} IconCacheHeader;
+
+static char *icon_cache_path(const char *icon_path, int w, int h) {
+  uint32_t hash = 5381;
+  for (const char *p = icon_path; *p; p++)
```
DaveDavenport: Please add curly brackets '{' '}'

source/rofi-icon-fetcher.c
```diff
@@ -50,10 +51,108 @@
...
+    hash = ((hash << 5) + hash) ^ (uint8_t)*p;
+  hash ^= (uint32_t)(w * 10007u ^ h * 100003u);
+  return g_strdup_printf("%s/" ICON_CACHE_DIR "/%08x.rofi", g_get_user_cache_dir(), hash);
```
DaveDavenport: below you use g_build_filename.
I think this should also be g_build_filename, instead of a g_strdup_printf

DaveDavenport:
Can you give a bit of a motivation/explanation for this MR?

As I don't directly see how this fixes the issue in the linked ticket.

The 'flickering' is caused by rofi having to redraw its icons because only on the 2nd drawing to the screen the scaling factor is known and it then has to reloads the icons (in the right resolution/size).
An icon cache does not directly solve this issues (might 'hide' the effect a bit on slower machines).

Zebra2711:
> Can you give a bit of a motivation/explanation for this MR?
> 
> As I don't directly see how this fixes the issue in the linked ticket.
> 
> The 'flickering' is caused by rofi having to redraw its icons because only on the 2nd drawing to the screen the scaling factor is known and it then has to reloads the icons (in the right resolution/size). An icon cache does not directly solve this issues (might 'hide' the effect a bit on slower machines).

Yeah, the first run without a cache will still flicker. I save the cache icons at the same size they load at, so it should reduce flicker once everything is cached since it won't need to call `gdk_pixbuf_new_from_file_at_scale`

DaveDavenport:
I suspect the flicker is still there, it is just less visible. All the loading logic is unchanged, beside the speed of loading the icon.

Let me play with the patch.

Zebra2711:
> I suspect the flicker is still there, it is just less visible. All the loading logic is unchanged, beside the speed of loading the icon.
> 
> Let me play with the patch.

It will still happen with big icons, i suppose :)

DaveDavenport:
I am on X11 (3xfhd) and a fast machine with a fast disk I don't notice a lot of speedup. But this might not be a fair check.

Will check later on a slower machine with 4k screen and wayland. 

I do notice that the cache is growing in size pretty quick, I see there is no limit per-image size. 

```
3.0M Mar 23 12:18 ffca750d.rofi
3.1M Mar 23 12:18 de31867e.rofi
```
If I use the fullscreen-preview theme.

If we add this, we should probably add some measures to make sure the cache does not explode. I've seen several rofi scripts that preview a large images..

@lbonn  what do you think?

Zebra2711:
> If we add this, we should probably add some measures to make sure the cache does not explode. I've seen several rofi scripts that preview a large images..

Sure that i need to add an option to disable the cache, or skip saving for images > 512x512/256x256/128x128px?

PR #2278, opened 2026-03-25
fix: enable input method switching under Wayland

breakatop:
1. Release the shortcut Control+Space for switching input methods; change the original Control+Space to Alt+Return.  
2. Implement pre‑edit, commit, and delete, allowing fcitx5 to function properly.


lbonn:
I've tested it and there is a bad interaction with the `-click-to-exit` option (on by default) we just merged in #2272, probably due to the additional layer surface.

Input method switching appears on the top-left of the screen instead of where the text input is. Problem disappears with `-no-click-to-exit`

But it's not exactly a problem with these changes I suppose, maybe acceptable as a first solution

source/wayland/display.c
```diff
@@ -1256,7 +1257,20 @@ static void text_input_enter(void *data, struct zwp_text_input_v3 *text_input,
   zwp_text_input_v3_commit(text_input);
 }
 static void text_input_leave(void *data, struct zwp_text_input_v3 *text_input,
-                             struct wl_surface *surface) {
```
lbonn, line 1259: What happened with these indentation changes? Better leave them aligned as clang-format does. Also in text_input_commit_string

source/keyb.c
```diff
@@ -191,7 +191,7 @@ ActionBindingEntry rofi_bindings[] = {
      .comment = "Go to the last entry"},
     {.id = ROW_SELECT,
      .name = "kb-row-select",
-     .binding = "Control+space",
+     .binding = "Alt+Return",
```
lbonn, line 194: To change in `doc/rofi-keys.5.markdown`.

Maybe Alt+space is closer to the original binding, not sure what's the most intuitive here.

PR #2282, opened 2026-04-07
Add setting to include directories in recursivebrowser

jacobwinters:
I'd prefer to be able to open directories, not just files, from recursivebrowser. My use case is making a giant combi config that can open anything on the system, sort of like Spotlight on a Mac.

I wrote a patch for my own use and have been using it for several months. #2212 makes it look like there's interest in upstreaming the functionality.

### DIFF
--- a/doc/rofi.1.markdown
+++ b/doc/rofi.1.markdown
@@ -840,6 +840,8 @@ configuration {
       filter-regex: "(.*cache.*|.*\.o)";
       /** command */
       command: "xdg-open";
+      /** show directories in results */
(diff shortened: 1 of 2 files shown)


kbuckleys:
I remember making a request for exactly this a while back. Hope it gets merged. Thank you so much.

yoru-kurai:
sorry to hijack this merge request, do you know how difficult it would be to also have a setting to include hidden items? just from looking at the code this is difficult for me to estimate. If you think it should be fairly simple, i will look into how to do it!

Issue #2143, opened 2025-06-05
[REQUEST] combi live input to other plugins/modes


zspher:
### Before opening a feature request

- [x] I checked the *next* branch to see if the feature has already been implemented
- [x] I searched existing reports to see if it is already requested.

### What is the user problem or growth opportunity you want to see solved?

Hi,

I'm trying to create a calc plugin for rofi in rust with rink. As a part of the plugin, I update the top most entry with the current user input every time a character is input with `rofi_view_reload()` on `_preprocess_input`.

I have tested this with  my mode's `_preprocess_input` but  `_preprocess_input` is never called by combi.

I am trying to integrate the plugin with combi mode. Can a feature to combi that allows a reload of entries, in general or when a plugin's bang prefix is triggered?

### How do you know that this problem exists today? Why is this important?

This feature allows immediate results for users who use plugins like calculators.

### Who will benefit from it?

plugin creators that need constantly updated content on user input

### Rofi version (rofi -v)

1.7.9+wayland1

### Configuration

https://gist.github.com/zspher/103fb661067d0d6345dc22ecc55116ad

zspher:
I have found a partial solution for this on my end - reload entries on `result` `MENU_OK & MENU_CUSTOM_ACTION`.

Issue #1934, opened 2024-01-10
Ability to *replace* the list in combi mode via a script


schaefsteven:
### Before opening a feature request

- [X] I checked the *next* branch to see if the feature has already been implemented
- [X] I searched existing reports to see if it is already requested.

### What is the user problem or growth opportunity you want to see solved?

Scripts that return new options (such as confirmation dialogs) do not replace the entire list of options in combi mode. They only replace their own options.

### How do you know that this problem exists today? Why is this important?

I have tested and looked for a solution and can not find any.

### Who will benefit from it?

Users who wish to have a lot of functionality in combi mode, which reduces the need to remember multiple keyboard shortcuts

### Rofi version (rofi -v)

1.7.3

### Configuration

https://gist.github.com/schaefsteven/3ebbe2a001ecebed7af23cdd36eeffe4

### Additional information

If a script returns new options to rofi, rofi (in single-mode operation) replaces the current list of options with the new list. 
In combi mode, the other modi's options remain in the list. 

In my case, I have combi mode with drun and a Power Management script, which asks for confirmation before executing its commands, but the confirmation options appear at the end of the options list (off the screen), which is not the desired function.

schaefsteven:
Note: one workaround for this is to run a new instance of rofi instead of returning new items to the existing instance. However this makes the code of the script much more clunky (the first instance calls your script again, which needs to parse the argument, and then run rofi, running your script a third time with another more different argument), causes the window to blink closed for a moment (not a big deal, but still worth mentioning), and just overall feels like the wrong answer.

DaveDavenport:
I don't understand this request..  You want want to combine the lists of multiple modes, this is what combi mode does and what you indicate it does, but also not?


It sounds more liike you do not want to use combi mode, but open multiple modes at the same time (under one keybinding), something rofi can do.

schaefsteven:
> I don't understand this request.. You want want to combine the lists of multiple modes, this is what combi mode does and what you indicate it does, but also not?

Yes, when I call Rofi with my key bind, I want it to show multiple modes, but then when I use that list to call the option from one of my modes, I want what that mode returns to be the only thing listed. 

I guess I don't really understand the use case of combi mode with its current implementation **when calling a mode/script that returns more options.** Why would I call a mode to get more options, but then choose something else entirely? That would be like driving to the Pizza shop but then not buying a pizza and instead leaving, going to the laundry mat, and doing my laundry. 

[...]

PR #2288, opened 2026-05-04
combi: makes combi compatible with `mode_preprocess_input`

### LINKED ISSUE #33: Not working in combi mode
Hi,
the plugin works for me if I launch rofi with the supplied command. But I want it to be available in combi mode. I tried adding it to the list of modes but it doesn't work. 
If I type something like 5*3 I get one entry saying "calc Add to history" but no result.
Do you know what the problem is?

My config file:
```
rofi.modi:                           combi

rofi.combi-modi:                     window,drun,ssh,calc
```

mtoons:
This PR allows modes `mode_preprocess_input` functions to work in `combi` mode.

This solves #2143 and probably #1934 but I haven't tested it with this person script.

Would also fix [this popular plugin issue](https://github.com/svenstaro/rofi-calc/issues/33).

### DIFF
--- a/source/modes/combi.c
+++ b/source/modes/combi.c
@@ -300,8 +300,13 @@ static cairo_surface_t *combi_get_icon(const Mode *sw, unsigned int index,
 
 static char *combi_preprocess_input(Mode *sw, const char *input) {
   CombiModePrivateData *pd = mode_get_private_data(sw);
+  char *input_preprocessed = g_strdup(input);
   for (unsigned i = 0; i < pd->num_switchers; i++) {
     pd->switchers[i].disable = FALSE;
+    char *input_preprocessed_old = input_preprocessed;
+    input_preprocessed =
+        mode_preprocess_input(pd->switchers[i].mode, input_preprocessed);
+    g_free(input_preprocessed_old);
   }
   if (input != NULL && input[0] == '!') {
     // Implement strchrnul behaviour.
--- a/source/modes/combi.c
+++ b/source/modes/combi.c
@@ -332,7 +332,7 @@ static char *combi_preprocess_input(Mode *sw, const char *input) {
       return g_strdup(eob + 1);
     }
   }
-  return g_strdup(input);
+  return g_strdup(input_preprocessed);
 }
 
 Mode combi_mode = {.name = "combi",
--- a/source/modes/combi.c
+++ b/source/modes/combi.c
@@ -312,6 +312,7 @@ static char *combi_preprocess_input(Mode *sw, const char *input) {
     }
     ssize_t bang_len = g_utf8_pointer_to_offset(input, eob) - 1;
     if (bang_len > 0) {
+      char *stripped_input = g_strdup(eob + 1);
       for (unsigned i = 0; i < pd->num_switchers; i++) {
         const char *mode_name = mode_get_name(pd->switchers[i].mode);
         size_t mode_name_len = g_utf8_strlen(mode_name, -1);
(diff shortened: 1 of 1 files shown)


DaveDavenport:
Took a very quick look.

This leaks memory and I don't think it implements the preprocessing behaviour correctly.

Now the input is pre-processed by every mode but the result is not used/returned. 


The idea behind the return off the pre-process callback is that if you  f.e. strip markup information that should not be used for searching or in combi to strip the 'bang' so that does not make a match on rest of the search (the result is used to create the search patterns).


tbh there might not be an easy way to implement this fully 'correct'.

mtoons:
> This leaks memory and I don't think it implements the preprocessing behaviour correctly.
>
> Now the input is pre-processed by every mode but the result is not used/returned.

I'm absolutely sorry, mistake on my part I simply forgot to return the modified preprocessed input.


> I don't think it implements the preprocessing behaviour correctly.
>
> The idea behind the return off the pre-process callback is that if you f.e. strip markup information that should not be used for searching or in combi to strip the 'bang' so that does not make a match on rest of the search (the result is used to create the search patterns).
>
>tbh there might not be an easy way to implement this fully 'correct'.

I'm definitely not a experienced contributor but I don't see why this implementation wouldn't be correct ?
Maybe I should apply the modes preprocess _after_ striping markup implementation, but even then it doesn't seem to hard to me.

DaveDavenport:
You still leak memory,  you added a g_strdup(), but not a matching free..
(last input_preprocessed  is still leaked).

The problem is that to do this correctly, the preprocessing using mode X, should be done only for entries from mode X and generate a search pattern only for entries from X.
Chaining pre-processing is not guaranteed giving the same behaviour, both in search result as in any other side-effects that pre-processing does.
This is currently not possible to implement in the current plugin api, hence my 'there might nog be an easy way to implement this correctly'.

   
I do agree, it  probably *mostly* work fine and people won't notice. But it can result in odd unexplained behaviour that will be confusing.

If we ignore this; I would however in the combi mode,  First do the !bang detection, then based on the bang, only pre-process with the matched switcher. With no bang, maybe chain it? (that still leaves the above issue).

DaveDavenport:
That last commit still leaks memory if there is a !bang.

mtoons:
I will try to implement it with bang tomorrow as you suggested.

Still I think it's better to have it working in some scenarios than never, chains seems natural to me even if it breaks in some edge cases. Some pretty popular plugin like rofi-calc hard-depend on `mode_preprocess_input` and I think it's a miss for it to not work in combi.

DaveDavenport:
I agree. sorry if that was not clear in my last (long) message.

mtoons:
Oh sorry I didn't catch that, I'm not great at English.

My last commit should implement chaining different modes `preprocess_input` with and without bang.
I haven't tested it properly yet, will do this afternoon.

If you think it's better I can add a `preprocessed_input` field in the `CombiModePrivateData` struct (or `CombiMode` I don't really now which one fits best) populate it in `combi_preprocess_input` and then use it in `combi_mode_match`.

DaveDavenport:
> If you think it's better I can add a preprocessed_input field in the CombiModePrivateData struct (or CombiMode I don't really now which one fits best) populate it in combi_preprocess_input and then use it in combi_mode_match.

I don't understand why you need it in `combi_mode_match`? . `rofi_int_matcher **tokens` is based on the value returned from _preprocess.

mtoons:
I was thinking it was necessary to have a different preprocessed input for each plugin.
I could probably do something with an escape value in the `combi_preprocess_input` return but that would be kind of hacky.

DaveDavenport:
As I mentioned earlier,  in the current setup this is not really supported.

Anything would be really hacky and probably causes performance issues. You would need to recalculate the regex patterns depending on row, or keep copies around (running in possible race-conditions with multi-threading) ). I don't think I will accept a patch that does that.
To do this correct would require a significant restructure/rewrite of the matching code.

mtoons:
Yeah sorry I think I underestimated the performance impact of `helper_tokenize()` I was thinking a plugin so "meta" is bound to contain some hacky stuff.

Anyway I'll test the chain method and reopen this PR once it works later today.

mtoons:
Yep it seems to be working as intended, no issues in my testing.

DaveDavenport:
Thanks!

mtoons:
Thank you for your patience!

Issue #2231, opened 2025-11-10
[BUG] window-command not working under Wayland (niri)


Lodifice:
### Rofi version (rofi -v or git commit in case of build issue)

2.0.0

### Configuration

https://gist.github.com/Lodifice/ed1ac229644f9d05717030027d214c40

### Theme

https://gist.github.com/Lodifice/dcea5f9f08e8e4fb4bd28fbb0ff35a7e

### Launch command

rofi -show window -mode window -window-command "notify-send rofi {window}"

### Step to reproduce

Just run the launch command from a terminal (optionally skip -window-command since it is also in the config). Then press Shift+Enter on any entry.

### Expected behavior

A notification pops up.

### Actual behavior

The window manager (niri) switches to the selected window.

### Additional information

I've tried different commands and different key bindings for kb-accept-alt. It's always the same, as if the normal accept was hit. However, under Xorg (dwm) it works as intended.

### I've checked if the issue exists in the latest stable release

- [x] Yes, I have checked the problem exists in the latest stable version

lbonn:
See also the initial PR for wayland window mode: https://github.com/lbonn/rofi/pull/58.

It was not done yet because we don't have an equivalent to the window id in X11 in this case, so it's less useful to compose with other tools and the place-holders for formatting the commands should be reworked and probably diverge between the two backends.

Lodifice:
Thanks for the pointer. So there is no concept of window id in wayland, but many compositors still use this concept. E.g. in niri window ids can be easily obtained by `niri msg windows`. Wouldn't it be possible to incorporate this into the place-holder?

I must admit that I have no idea how rofi queries the compositor for the list of windows. If that's a general compositor-transparent way that just doesn't expose window ids, then I'd assume would unfeasible to add a specific way to obtain the window ids for every particular compositor.

LRitzdorf:
Hopefully this isn't a thread hijack — I'm running into the same issue under Hyprland, which is surprising, since I thought I'd tested a few months ago and found that `{window}` *was* valid (and resolved to the app's PID).

OP didn't include any debug logs, so I can't say for sure that their root cause is the same as mine, but I'm getting a parser error, of all things:
```ini
[XrmOptions]: str: "configuration { window { command: hyprctl notify 1 5000 0 asdf; } }"
[default]: Parsing string: 'configuration { window { command: hyprctl notify 1 5000 0 asdf; } }'
[Parser]: Failed to parse theme:
[...]

PR #2291, opened 2026-05-14
Additional feature support for WaylandWindow

LRitzdorf:
Adds additional feature support for `window` mode under Wayland, bringing it closer to feature parity with the X11 version. Key changes are:
- [X] Handles `accept-custom` by running the entered command
- [X] Respects the `window.close-on-delete` config setting
- [X] Fixes self-contradictory keybind documentation
- [x] ~~Fills in the `{window}` placeholder in `window-command` with something reasonable~~ See #2295

Still a work in progress with regard to that last point. I'm completely new to Wayland protocol interactions, and while I think I've got the gist of it from looking at the existing code that speaks `wlr-foreign-toplevel-management-unstable-v1`, I'll definitely have questions for someone with more experience here.

### DIFF
--- a/doc/rofi.1.markdown
+++ b/doc/rofi.1.markdown
@@ -982,21 +982,25 @@ configuration {
 
 ### window
 
-Show a list of all the windows and allow switching between them. Pressing the
-`delete-entry` binding (`shift-delete`) will close the window. Pressing the
-`accept-custom` binding (`control-enter` or `shift-enter`) will run a command
-on the window. (See option `window-command` );
+Show a list of all the windows and allow switching between them.
+Pressing the `delete-entry` binding (`shift-delete`) will close the window.
+Pressing the `accept-alt` binding (`shift-enter`) will run a command on the
+window. (See option `window-command` );
 
-If there is no match, it will try to launch the input.
+If there is no match, or if `accept-custom` (`control-enter`) is pressed, it
+will try to launch the input.
 
 ### windowcd
 
+Same as the **window** mode, but lists only windows on the current desktop.
 Shows a list of the windows on the current desktop and allows switching between
-them. Pressing the `delete-entry` binding (`shift-delete`) will kill the
-window. Pressing the `accept-custom` binding (`control-enter` or `shift-enter`)
-will run a command on the window. (See option `window-command` );
+them.
+Pressing the `delete-entry` binding (`shift-delete`) will kill the window.
+Pressing the `accept-alt` binding (`shift-enter`) will run a command on the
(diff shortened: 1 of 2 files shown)


LRitzdorf:
I'm kind of just throwing this out here for now, in the hopes that someone with Wayland experience will weigh in (@lbonn perhaps? Apologies for the ping).

It'd be really nice if we could support `window-command` under Wayland, but as mentioned in https://github.com/davatorium/rofi/issues/2231#issuecomment-3536328938 and lbonn#58, figuring out which placeholders to have for this is kind of annoying.

One candidate is `ext-foreign-toplevel-list-v1`'s [`ext_foreign_toplevel_handle_v1::identifier`](https://wayland.app/protocols/ext-foreign-toplevel-list-v1#ext_foreign_toplevel_handle_v1:event:identifier), which is "a stable identifier for a toplevel," and notably is more unique than just app_id plus title (which could easily match multiple windows). I'm completely new to Wayland protocol interactions, but speaking a new protocol doesn't seem like a massive hurdle.

What I'm more worried about is figuring out which handle (from `wlr-foreign-toplevel-management-unstable-v1`) corresponds to which identifier; i.e. identifying the same window as reported by two different protocols. Both protocols do provide title and app_id, but again, that's not enough to uniquely identify a window.

Thoughts?

---
There *is* a solution to this for Hyprland specifically, since it has [`hyprland-toplevel-mapping-v1`](https://wayland.app/protocols/hyprland-toplevel-mapping-v1). This could be used either to identify corresponding ext and wlr toplevels, or just to get the Hyprland "address" for a wlr toplevel (which could then be used in window-command). Obviously though, this wouldn't work on any other compositors, so is probably(?) not worth implementing here.

lbonn:
@LRitzdorf thanks! So, two things:

- commits 1 and 3 are worthwhile on their own, if there is still some open question for the last part it's probably a good idea to open a separate PR for them

- for the launching command part, the minimum use case we can merge is probably anything that can be used from an external script. If we supply ext_foreign_toplevel_handle_v1::identifier and it's possible in some DM to map it back to some other identifiers, maybe it would count :). So, if you can demonstrate a script (even not completely portable) that does something useful from there, I would count it fair game

LRitzdorf:
Sounds good! I'll mark this as ready for review, then, and open a new WIP PR for window-command stuff.

Also just to clarify, commit 2 (which allows launching entered commands in window mode) is completely separate from the `window-command` and Wayland identifier discussion that we still need to figure out. I.e. nothing in this PR is window-command-related

PR #2293, opened 2026-05-19
Fix -no-show-icons setting

ghost:
When using a theme that specifies children for the `element` widget the `-no-show-icons` setting no longer takes effect. Instead of just handling the default case, this commit filters out `element-icon` widgets if `-no-show-icons` is set.

### DIFF
--- a/source/widgets/listview.c
+++ b/source/widgets/listview.c
@@ -172,8 +172,10 @@ static void listview_set_state(_listview_row r, TextBoxFontType tbft) {
 static void listview_add_widget(listview *lv, _listview_row *row, widget *wid,
                                 const char *label) {
   if (strcasecmp(label, "element-icon") == 0) {
-    row->icon = icon_create(WIDGET(wid), "element-icon");
-    box_add((box *)wid, WIDGET(row->icon), FALSE);
+    if (config.show_icons) {
+      row->icon = icon_create(WIDGET(wid), "element-icon");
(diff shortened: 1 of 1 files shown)


lbonn (approved):
looks good to me. I'll leave @DaveDavenport to double check before merging

DaveDavenport:
This is an option I would like to deprecate.  Showing/Not showing icons is a theme option and should not be a separate dedicated commandline option.
This is mostly here for 'compatibility' reason. 

To disable an icon you can do at end of config/theme file:

```css
element-icon {
    enabled: false;
}
```

or on commandline you can pass it via `-theme-str` or `-theme+element-icon+enabled false`

DaveDavenport:
I think this change would break all themes that have icons shown by default, as now suddenly it requires the `-show-icons` option.  I don't think that is acceptable as it would confuse tons of users.

DaveDavenport:
Updated documentation to clarify that this options applies to the default theme.

ghost:
Sorry, I didn't realize that this would affect so many configurations.
Thanks for the `-theme+element-icon+enabled false` hint! That works great for me.

Closing this PR

DaveDavenport:
Took me a second to realize as well. 
(one reason why I don't like this 'deprecated' option, its from when we did not have advanced themeing, only coloring, so to keep option it modifies the default setup of a listview element.)

I should deprecate the options, but people get upset when options get removed/changed.

PR #2295, opened 2026-05-20
Window: support `window-command` under Wayland

### LINKED ISSUE #2231: [BUG] window-command not working under Wayland (niri)

LRitzdorf:
Continuing discussion from #2291 in a dedicated PR.
~~**NOTE:** This PR is currently stacked with the above, so its diff is going to look weird. I'll rebase this once #2291 is merged.~~ Done!

---

It'd be really nice if we could support `window-command` under Wayland, but as mentioned in https://github.com/davatorium/rofi/issues/2231#issuecomment-3536328938 and lbonn#58, figuring out which placeholders to have for this is kind of annoying.

One candidate is `ext-foreign-toplevel-list-v1`'s [`ext_foreign_toplevel_handle_v1::identifier`](https://wayland.app/protocols/ext-foreign-toplevel-list-v1#ext_foreign_toplevel_handle_v1:event:identifier), which is "a stable identifier for a toplevel," and notably is more unique than just app_id plus title (which could easily match multiple windows). I'm completely new to Wayland protocol interactions, but speaking a new protocol doesn't seem like a massive hurdle.

What I'm more worried about is figuring out which handle (from `wlr-foreign-toplevel-management-unstable-v1`) corresponds to which identifier; i.e. identifying the same window as reported by two different protocols. Both protocols do provide title and app_id, but again, that's not enough to uniquely identify a window.

Thoughts?

Fixes #2231

### DIFF
--- a/source/modes/wayland-window.c
+++ b/source/modes/wayland-window.c
@@ -170,15 +170,29 @@ static void wlr_foreign_toplevel_handle_close(WlrForeignToplevelHandle *self) {
 
 static void ext_foreign_toplevel_handle_done(
     void *data, G_GNUC_UNUSED struct ext_foreign_toplevel_handle_v1 *handle) {
+  ExtForeignToplevelHandle *self = (ExtForeignToplevelHandle *)data;
+
+  g_debug("ext window %p id=%s identifier=%s", (void *)self,
+          self->app_id, self->identifier);
 }
 
 static void ext_foreign_toplevel_handle_closed(
     void *data, G_GNUC_UNUSED struct ext_foreign_toplevel_handle_v1 *handle) {
+  ExtForeignToplevelHandle *self = (ExtForeignToplevelHandle *)data;
+
+  /* the handle is inert and will receive no further events */
+  self->view->ext_toplevels = g_list_remove(self->view->ext_toplevels, self);
+  ext_foreign_toplevel_handle_free(self);
 }
 
 static void ext_foreign_toplevel_handle_identifier(
     void *data, G_GNUC_UNUSED struct ext_foreign_toplevel_handle_v1 *handle,
     const char *identifier) {
+  ExtForeignToplevelHandle *self = (ExtForeignToplevelHandle *)data;
+  if (self->identifier) {
+    g_free(self->identifier);
+  }
+  self->identifier = g_strdup(identifier);
 }
 
 static void ext_foreign_toplevel_handle_title(
@@ -191,6 +205,11 @@ static void ext_foreign_toplevel_handle_title(
 static void ext_foreign_toplevel_handle_app_id(
     void *data, G_GNUC_UNUSED struct ext_foreign_toplevel_handle_v1 *handle,
     const char *app_id) {
+  ExtForeignToplevelHandle *self = (ExtForeignToplevelHandle *)data;
+  if (self->app_id) {
+    g_free(self->app_id);
+  }
+  self->app_id = g_strdup(app_id);
 }
 
 static struct ext_foreign_toplevel_handle_v1_listener
@@ -204,12 +223,23 @@ static struct ext_foreign_toplevel_handle_v1_listener
 static ExtForeignToplevelHandle *
 ext_foreign_toplevel_handle_new(struct ext_foreign_toplevel_handle_v1 *handle,
                                 WaylandWindowModePrivateData *view) {
-  return NULL;
+  ExtForeignToplevelHandle *self =
+      (ExtForeignToplevelHandle *)g_malloc0(sizeof(ExtForeignToplevelHandle));
+
+  self->handle = handle;
+  self->view = view;
+  ext_foreign_toplevel_handle_v1_add_listener(
+      handle, &ext_foreign_toplevel_handle_listener, self);
+  return self;
 }
 
 static void ext_foreign_toplevel_list_toplevel(
     void *data, G_GNUC_UNUSED struct ext_foreign_toplevel_list_v1 *list,
     struct ext_foreign_toplevel_handle_v1 *toplevel) {
+  WaylandWindowModePrivateData *pd = (WaylandWindowModePrivateData *)data;
+
+  ExtForeignToplevelHandle *handle = ext_foreign_toplevel_handle_new(toplevel, pd);
+  pd->ext_toplevels = g_list_prepend(pd->ext_toplevels, handle);
 }
 
 static void ext_foreign_toplevel_list_finished(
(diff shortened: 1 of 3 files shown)


LRitzdorf:
From https://github.com/davatorium/rofi/pull/2291#issuecomment-4487309271:
> for the launching command part, the minimum use case we can merge is probably anything that can be used from an external script. If we supply ext_foreign_toplevel_handle_v1::identifier and it's possible in some DM to map it back to some other identifiers, maybe it would count :). So, if you can demonstrate a script (even not completely portable) that does something useful from there, I would count it fair game

Okay, nice, that's a lower barrier to entry than I was expecting. I've kind of got three different options here:
1. With no additional protocols needed, we could make the app_id and title available for use in `window-command`. This would be easy and work under every compositor we already support, but limited in that it can't uniquely identify windows (e.g. two `kitty` instances with a title of `~`).
2. We could use `ext-foreign-toplevel-list-v1` to obtain a list of (app_id, title, identifier) tuples, and make the "identifier" field available in `window-command`. Problem is, mapping these to the handles we already get from `wlr-foreign-toplevel-management-unstable-v1` is tricky. The only information they share is app_id and title, but again, those aren't enough to uniquely map an "ext" toplevel to the corresponding "wlr" toplevel.
    - As a heuristic, maybe we use the order that toplevels are received? E.g. when matching an ext toplevel to two possible wlr toplevels (with the same app_id and title), we choose the wlr toplevel that was received earliest, under the assumption that the compositor will iterate over windows in the same order as it sends wlr and ext events. **Testing this is probably my next step.**
4. We could dive into compositor-specific protocols, but that seems like a can of worms we may not want to open. Although, for now, I only see two protocols that could give us some form of window ID: [`hyprland-toplevel-mapping-v1`](https://wayland.app/protocols/hyprland-toplevel-mapping-v1#hyprland_toplevel_window_mapping_handle_v1:event:window_address) and [`treeland-foreign-toplevel-manager-v1`](https://wayland.app/protocols/treeland-foreign-toplevel-manager-v1#treeland_foreign_toplevel_handle_v1:event:identifier).
    - The Hyprland protocol lets us get the Hyprland "window address" for either a wlr toplevel or an ext toplevel. This, on its own, would let us solve option 2 above (by mapping wlr to ext toplevels), and we could then provide the ext toplevel identifier to `window-command`.
  ...this doesn't actually help us much, though, because what's the point of using a nice generic ext identifier when we can only map it correctly under Hyprland? We might as well just use the Hyprland-specific window address, and not bother speaking the ext toplevel protocol.
    - The Treeland protocol gives us access to (app_id, title, identifier, PID) tuples, but those are subject to the same problem as option 2 above. We don't have a proper way to figure out which wlr toplevel a given Treeland toplevel corresponds to.

LRitzdorf:
Working on implementing item 2 above, and it looks like (at least on Hyprland) the ext and wlr toplevel protocols give different window titles for some reason. Specifically, the ext protocol gets sent the *initial* title, while the wlr one gets the *current* title. That means we can't identify corresponding ext and wlr toplevels using both class and title; we have to use *just* the class. Grr.

Good news though, it looks like (again on Hyprland), ext and wlr toplevels are received in a consistent and corresponding order each time rofi launches. That sequencing, plus window classes, should be enough to do that wlr-to-ext mapping, so we can attach an "identifier" (from the ext protocol) to each wlr toplevel.

Should have a working version soonish!

LRitzdorf:
Okay cool, this works, at least under Hyprland! I can now bring a window onto the current workspace via a combination of this PR and:
```css
window-command: "hyprctl eval 'for _,w in ipairs(hl.get_windows()) do if w.stable_id == 0x{window} then hl.dispatch(hl.dsp.window.move({ window = w, workspace = \"+0\" })) return end end'";
```
...which is slightly cursed because Hyprland doesn't let you directly select windows by identifier, but works nonetheless. (Edit: working on support for this in hyprwm/Hyprland#14984)

For now this is still stacked on top of #2291, but once that's merged, this is ready to go as well!

lbonn:
Thanks!

LRitzdorf:
Oop, used `bool` instead of `gboolean` there. Fixed now

LRitzdorf:
With the `stableid:` window selector just merged in hyprwm/Hyprland#14984 (shipping in ~~0.55.5~~ 0.56), this PR and
```css
window-command: "hyprctl dispatch 'hl.dsp.window.move({ window = \"stableid:{window}\", workspace = \"+0\" })'";
```
are all that's needed to act on a window via Rofi!

source/modes/wayland-window.c
```diff
@@ -393,7 +550,56 @@ static unsigned int wayland_window_mode_get_num_entries(const Mode *sw) {
...
+  // match up wlr to ext toplevels, and set their identifiers
+  if (toplevel->identifier == NULL) {
+    /* NOTE: This assumes toplevels of the same class are received in the same
+     * order via both the wlr and ext protocols! i.e. if we get the order
+     * [firefox, kitty1, kitty2] via wlr and [kitty1, firefox, kitty2] via ext,
+     * that's fine, because the two kitty instances are in the same order.
+     * However, getting [kitty2, firefox, kitty1] via ext would break our
+     * mapping of ext to wlr toplevels, so a window-command on kitty1 would
+     * target kitty2 instead.
+     */
```
lbonn, line 567: Do you know where this undocumented behavior holds?
It was tested on hyprland and sway current versions?

LRitzdorf: Yep! I developed this under Hyprland 0.55, and just now tested under Sway 1.12 and Niri 26.04, where it seems to work reliably as well.

That said, this is absolutely a hack, since nothing in the Wayland spec makes any guarantees about event ordering.

lbonn:
Very cool, I think it's ok if it's a bit hacky for now :). It touches quite a specialized part of the code.

PR #2296, opened 2026-05-20
Add hover-select mode to listview widget

haxllo:
## Summary
- Add hover-select mode to listview: scroll events move viewport without changing selection, item under cursor is auto-selected
- Fix hover-select lag caused by redundant hover update in draw loop
- Map scrollbar to viewport offset in hover-select mode so it tracks position correctly

## Changes
- Add `hover_select` flag, mouse position tracking, viewport offset management
- Remove `listview_update_hover_from_mouse` from `listview_draw` (was causing feedback loop)
- Scrollbar uses viewport offset instead of selected item when hover-select is active

## Test plan
- [ ] `rofi -show drun -hover-select` — hover should be smooth, no lag
- [ ] Scroll to bottom — scrollbar should reach the end

### DIFF
--- a/source/view.c
+++ b/source/view.c
@@ -761,13 +761,6 @@ static void page_changed_callback(void) {
   rofi_view_workers_initialize();
 }
 
-static void hover_update_callback(void *udata) {
-  (void)udata;
-  if (config.hover_select && current_active_menu) {
-    rofi_view_ping_mouse(current_active_menu);
-  }
-}
-
 static void _rofi_view_reload_row(RofiViewState *state) {
   g_free(state->line_map);
   g_free(state->distance);
@@ -1745,7 +1738,6 @@ static void rofi_view_add_widget(RofiViewState *state, widget *parent_widget,
     box_add((box *)parent_widget, WIDGET(state->list_view), TRUE);
     listview_set_scroll_type(state->list_view, config.scroll_method);
     listview_set_hover_select(state->list_view, config.hover_select);
-    listview_set_hover_update_callback(state->list_view, hover_update_callback, NULL);
     listview_set_mouse_activated_cb(
         state->list_view, rofi_view_listview_mouse_activated_cb, state);
 
--- a/source/view.c
+++ b/source/view.c
@@ -761,6 +761,13 @@ static void page_changed_callback(void) {
   rofi_view_workers_initialize();
 }
 
+static void hover_update_callback(void *udata) {
+  (void)udata;
+  if (config.hover_select && current_active_menu) {
+    rofi_view_ping_mouse(current_active_menu);
+  }
+}
+
 static void _rofi_view_reload_row(RofiViewState *state) {
   g_free(state->line_map);
   g_free(state->distance);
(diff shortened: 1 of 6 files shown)


source/view.c
```diff
@@ -761,6 +761,13 @@ static void page_changed_callback(void) {
   rofi_view_workers_initialize();
 }
 
+static void hover_update_callback(void *udata) {
+  (void)udata;
```
DaveDavenport: We use G_GNUC_UNUSED for this. 
https://docs.gtk.org/glib/macros.html

source/widgets/listview.c
```diff
@@ -459,13 +471,34 @@ static void listview_draw(widget *wid, cairo_t *draw) {
...
-  scrollbar_set_handle_length(lv->scrollbar, lv->cur_columns * lv->max_rows);
-  if (lv->reverse) {
-    scrollbar_set_handle(lv->scrollbar, lv->req_elements - lv->selected - 1);
+  if (lv->hover_select) {
+    unsigned int max_offset = lv->req_elements > lv->max_elements
+                                  ? lv->req_elements - lv->max_elements
+                                  : 0;
+    scrollbar_set_max_value(lv->scrollbar, max_offset);
+    scrollbar_set_handle_length(lv->scrollbar, 1);
+    scrollbar_set_handle(lv->scrollbar, lv->viewport_offset);
```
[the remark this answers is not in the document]
haxllo: Thanks for the thorough analysis. I looked into this and decided not to change it because:

1. When `req_elements <= max_elements`, all items fit on screen — there's nothing to scroll, so the scrollbar is cosmetic only. The `handle_length=1` renders a minimal handle which is fine for this case.

2. The scrollbar widget internally guards against division by zero — `scrollbar_draw()` uses `(length - 1)` only when length > 1.

3. For the normal case (`max_offset > 0`), `handle_length=1` means the handle represents 1 visible page worth of items, which matches the viewport-based model where each scroll step is one item.

That said, if you'd prefer I adjust the handle sizing, happy to follow your guidance.

source/view.c
haxllo: Fixed — removed the callback entirely as it was dead code (hover re-evaluation is already handled by `listview_update_hover_from_mouse` in the scroll handlers).

DaveDavenport:
I works well (did a quick test), except for clicking on the scrollbar, that is now broken?

haxllo:
> I works well (did a quick test), except for clicking on the scrollbar, that is now broken?

let me look  into it

haxllo:
> I works well (did a quick test), except for clicking on the scrollbar, that is now broken?

fixed it ,the issue was that clicking/dragging the scrollbar called listview_set_selected which only changed the selection index but never updated viewport_offset so the list appeared not to scroll. I added listview_scrollbar_scroll_to that properly updates viewport_offset, syncs positions, and redraws in hover-select mode

"><also bundled a few UX improvements while at it: jump on press instead of release, page up/down on track clicks, pressed-state handle color, auto-hide when everything fits, and a wider click target.

DaveDavenport:
Thanks!

haxllo:
> Thanks!

Would you consider merging it, or are there still issues that need to be addressed first?

DaveDavenport:
I've asked others to give it a quick test. To see if they have remarks.

lbonn:
With `-hover-select`, leaving the cursor in place and moving the selection with arrow keys does not auto-scroll anymore.

PR #2315, opened 2026-07-07
Fix null pointer deref and skip misformed entry

jakobwsmnn:
## Description
I’ve notice a potential null pointer deref when parsing a misformed ssh config. 
In [ssh.c](https://github.com/davatorium/rofi/blob/next/source/modes/ssh.c#L205), when the SSH hostnames(IPv6) get parsed, the code uses strchr() to find “]”. If the input is malformed (it exists a "[" but no "]"), then strchr will return NULL and thus NULL gets derefd which results in a segfault. 

## Solution:
Add a NULL check after strchr(). If it is null, skip the entry and continue with the remaining entries.


DaveDavenport:
Thanks.

PR #2316, opened 2026-07-09
Replace malloc and realloc with corresponding glibc function

jakobwsmnn:
## Description
The entire codebase uses g_malloc and g_realloc, but [rofi.c](https://github.com/davatorium/rofi/blob/next/source/rofi.c#L871) and [ssh.c](https://github.com/davatorium/rofi/blob/next/source/modes/ssh.c#L505) do not.
g_malloc terminates the program on allocation failure. malloc does not.

Is there a particular reason why malloc is used here instead of g_malloc?
If malloc should be used here, a NULL check should be added. 

## Solution:
Replace malloc/realloc with g_malloc/g_realloc in rofi.c and malloc with g_malloc0 in ssh.c.

### DIFF
--- a/source/modes/ssh.c
+++ b/source/modes/ssh.c
@@ -494,7 +494,7 @@ static SshEntry *get_ssh(SSHModePrivateData *pd, unsigned int *length) {
   path = g_build_filename(cache_dir, SSH_CACHE_FILE, NULL);
   char **h = history_get_list(path, length);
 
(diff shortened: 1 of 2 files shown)


DaveDavenport:
No particular reason, mostly habit. 

I've tried to avoid it, must have missed these.

DaveDavenport:
Thanks.

PR #2306, opened 2026-06-17
Add fzf-v2 sorting method

mil-ad:
`-sorting-method fzf` is only loosely inspired by fzf (see #533) and doesn't rank like it. For example, with a file containing:

```
⏳ hourglass not done; wip; in progress
🔏 locked with pen
🫣 face with peeking eye
```

searching `wip` with `-sorting-method fzf` puts the literal "wip" line **last**, while real `fzf --filter=wip` puts it **first**. The current scorer rewards matches spread across word boundaries over contiguous ones.

This PR adds a new `-sorting-method fzf-v2` that ports fzf's default scoring algorithm (FuzzyMatchV2 from junegunn/fzf). The existing `fzf` method is left unchanged.

- Smith-Waterman style DP using fzf's own weights for word-boundary, camelCase and consecutive matches, plus its gap penalties.
- Whitespace-separated terms are scored independently and summed, matching fzf's AND semantics.
- Honours `-normalize-match` (scores the accent-stripped text, like fzf's rune normalization).

Usage:

```
rofi -dmenu -matching fuzzy -sort -sorting-method fzf-v2
```

Known limitation: Tie-breaking is not implemented. When two entries score equally, fzf orders them by length/chunk; this falls back to input order. Scores themselves match fzf, so only the ordering within equal-score groups can differ.

PR #2335, opened 2026-09-16
Adding length tiebreak for fzf-v2 sorter

mil-ad:
Follow-up to #2306.

## Problem

`fzf-v2` scores rows but never breaks ties, and ties are the normal case: fzf's score describes only the matched region, so every line matching a term at a word boundary scores the same.

```
$ printf '%s\n' 'baked apple pie' 'green apple tart' 'candy apple cake' 'pine-apple juice' 'apple' | rofi -dmenu -sort -sorting-method fzf-v2 -filter apple
```

Four of those lines score an identical 140, so `apple` lands 4th instead of 1st. Real fzf puts it first.

## Fix

Sort on fzf's own key rather than the bare negated score. fzf compares a `[4]uint16` tuple (`Result.points`, `src/result.go`) from the most significant element down; for its default scheme that's the inverted, uint16-clamped score followed by the length in code points ignoring surrounding whitespace (`{byScore, byLength}` in `parseScheme`, `src/options.go`). Packing both into one integer reproduces the ordering.

The length field gets 15 bits rather than fzf's 16, so the widest key is exactly `G_MAXINT`. That keeps it in the existing `int` weight and lets `lev_sort()` go on subtracting keys without overflow. Costs nothing in practice: the scorer rejects anything over `FUZZY_SCORER_MAX_LENGTH`, so any row that matches is far below the cap.

Rows tying on both criteria are left to the sort, where fzf falls back to input order. `g_qsort_with_data()` is stable from glib 2.82; below that such a tie is unspecified, as it already is for the other sorting methods.

Scoped to `SORT_FZF_V2` — `normal` and `fzf` ordering is untouched, and the only change outside the new helper is one hunk in `filter_elements()`.

## Testing

22 new assertions in `helper-test.c`. Ordering checked against `fzf --filter --scheme=default` over 21 queries including multi-term ones; all match.


source/view.c
```diff
@@ -861,7 +885,9 @@ static gboolean rofi_view_refilter_real(RofiViewState *state) {
       j += states[i].count;
     }
     if (config.sort) {
-      g_qsort_with_data(state->line_map, j, sizeof(int), lev_sort,
+      g_qsort_with_data(state->line_map, j, sizeof(int),
+                        config.sorting_method_enum == SORT_FZF_V2 ? fzf_v2_sort
+                                                                  : lev_sort,
```
DaveDavenport: For readability can this not be an inline ternary operator and just an if/else ?

source/view.c
```diff
@@ -478,10 +500,12 @@ static void filter_elements(thread_state *ts,
...
           break;
         case SORT_FZF_V2:
-          /* The scorer returns a higher-is-better score; the sort orders by
-           * ascending distance, so negate it. */
-          t->state->distance[i] = -rofi_scorer_fzf_v2_evaluate(
-              t->pattern, t->plen, str, slen, t->state->case_sensitive);
+          /* Scores tie constantly, so sort on fzf's key (inverted score plus
+           * a length tiebreak) rather than on the score alone. */
+          t->state->distance[i] = rofi_scorer_fzf_v2_sort_key(
+              rofi_scorer_fzf_v2_evaluate(t->pattern, t->plen, str, slen,
```
DaveDavenport: can this be split, so its not nested calls.
(readability)

source/helper.c
```diff
@@ -1340,6 +1340,56 @@ int rofi_scorer_fzf_v2_evaluate(const char *pattern, glong plen,
...
+  if (val < 0) {
+    return 0;
+  }
+  return (guint16)val;
+}
+
+/** Code points in `str` ignoring surrounding whitespace; fzf's TrimLength(). */
+static guint16 rofi_scorer_fzf_v2_trim_length(const char *str) {
+  glong len;
+  gunichar *txt = g_utf8_to_ucs4_fast(str, -1, &len);
```
DaveDavenport: Not sure about performance?  but do we need the extra allocation here?  wouldn't a `g_utf8_next_char` and `g_utf8_get_char` work for this?

DaveDavenport: `g_utf8_prev_char` for going backwards

DaveDavenport:
Thanks!  looks ok to me.

(Don't use this matcher myself, but closer to fzf is better as that is what people are used too)

source/view.c
mil-ad: Done

source/view.c
mil-ad: Done

source/helper.c
mil-ad: Good call, done. g_utf8_next_char forward and g_utf8_prev_char backward,

mil-ad:
Thanks Dave! I have addressed your comments.

